PR #23, opened 2020-07-17
MB-2970 Connect name and address fields to approved data json

abbyoung:
## Description

* This PR replaces Faker's generated data for first name, last name, and street address, with MilMove's approved safe fake data.
* Data from spreadsheet is stored in `static/fake_data.json`

## Reviewer Notes

* The createMoveTaskOrder endpoint isn't dynamically creating data yet, so you won't be able to use it to verify
* Use GET /move-task-orders and review the newly added mto_shipments to see the fake data
  * Addresses -> street_address_1, etc.
  * Names -> Agents

## Setup

```sh
# in /mymove
make dev_db_e2e_populate
make server_run

# in /milmove-load-testing
make load_test_prime
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2970) for this change.

### DIFF
--- a/utils/fake_data.py
+++ b/utils/fake_data.py
@@ -87,7 +87,7 @@ class MilMoveProvider(DateProvider):
         Returns a safe street address as a string.
         """
         address = self.random_element(self.safe_data["addresses"])
-        return address["address"]
+        return address
 
 
 class MilMoveData:
--- a/utils/fake_data.py
+++ b/utils/fake_data.py
@@ -1,10 +1,10 @@
 # -*- coding: utf-8 -*-
 """ utils/fake_data.py is for Faker classes and functions to set up test data """
 import logging
+import json
 from typing import Optional
 from copy import deepcopy
 from datetime import datetime
-
 from faker import Faker
 from faker.providers.date_time import Provider as DateProvider  # extends BaseProvider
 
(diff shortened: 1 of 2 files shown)


utils/fake_data.py
```diff
@@ -41,6 +51,44 @@ def iso_date_time(self):
...
+    def safe_first_name(self):
+        """
+        Returns a safe first name as a string.
+        """
+        if not self.first_name_used:
+            return self.current_name["first_name"]
+        else:
+            self.set_safe_name()
+            self.first_name_used = True
+            return self.current_name["first_name"]
```
sandy-wright, line 72: I would refactor this method a little to avoid duplicate returns. Like so:

```python
if self.first_name_used:
    self.set_safe_name()

self.first_name_used = True
return self.current_name["first_name"]
```

And I would do the same for `safe_last_name`. And I think maybe setting the attributes `self.first_name_used` and `self.last_name_used` to True needs to happen regardless of which part of the `if` statement we're in? Although it looks like it's working this way too, but I'm not sure why 🤔

utils/fake_data.py
```diff
@@ -41,6 +51,44 @@ def iso_date_time(self):
...
+            self.set_safe_name()
+            self.last_name_used = True
+            return self.current_name["last_name"]
+
+    def safe_street_address(self):
+        """
+        Returns a safe street address as a string.
+        """
+        address = self.random_element(self.safe_data["addresses"])
+        return address
```
sandy-wright, line 90: You could skip a step and do:

```python
return self.random_element(self.safe_data["addresses"])
``` 

but it works either way!

PR #26, opened 2020-07-29
MB-3398 Support API: Fix load testing task for createMoveTaskOrder endpoint

sandy-wright:
## Description

This PR changes the hardcoded payload for the task `create_move_task_order` so that it functions with the changes after consolidating the DB. It also reworks the way overrides work so that it's a little more intuitive. There's also some general clean up and fixing of other issues (such as the `create_mto_shipment` task). 

## Reviewer Notes

You will need to checkout the `sw-mb-3398-fix-create-mto` branch in the `mymove` repo in order to test these changes. Here is the PR: https://github.com/transcom/mymove/pull/4515

To run this against the changes to the YAML for the support and prime APIs, you may need to change the hardcoded url in `PrimeAPIParser` and `SupportAPIParser` to use the correct branch as well (instead of master). You can reasonably test without doing this however, and it will let you see that some requests to `create_mto_shipment` will fail because they have too many agents!

## Setup

In your `mymove` project, checkout `sw-mb-3398-fix-create-mto` and run:

```sh
make db_dev_e2e_populate && make server_run
```

Next, in your `milmove_load_testing` directory, please follow the instructions in the README to set up your environment. Then run:

```sh
make load_test_prime
```

Or:

```sh
. .venv/bin/activate && locust -f locustfiles/prime.py --host local
```

Then navigate to http://localhost:8089/, choose a random number of users and a hatch rate, then run the load tests. Most responses for all endpoints should be successful.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3398) for this change.

### DIFF
--- a/utils/constants.py
+++ b/utils/constants.py
@@ -14,6 +14,8 @@ PRIME_CERT_KWARGS = {"cert": (LOCAL_MTLS_CERT, LOCAL_MTLS_KEY), "verify": False}
 ARRAY_MIN = 1
 ARRAY_MAX = 5
 
+ZERO_UUID = "00000000-0000-0000-0000-000000000000"
+
 
 class MilMoveEnv(ListEnum):
     LOCAL = "local"
--- a/utils/fields.py
+++ b/utils/fields.py
@@ -48,6 +48,9 @@ class BaseAPIField:
         :param kwargs:
         :return: any
         """
+        if (overrides := kwargs.get("overrides")) and self.name in overrides:
+            return overrides[self.name]
+
         return faker.get_fake_data_for_type(self.data_type)
 
 
@@ -64,6 +67,9 @@ class EnumField(BaseAPIField):
         :param kwargs:
         :return: str/int/element
         """
+        if (overrides := kwargs.get("overrides")) and self.name in overrides:
+            return overrides[self.name]
+
         return faker.get_random_choice(self.options)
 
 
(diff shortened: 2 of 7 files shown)


utils/fields.py
```diff
@@ -83,9 +89,22 @@ def generate_fake_data(self, faker, **kwargs):
...
+            if isinstance(override_array, list):
+                for item in override_array:
+                    kwargs["overrides"] = item
+                    fake_data.append(self.items_field.generate_fake_data(faker, **kwargs))
+
+                return fake_data
+
+            kwargs["overrides"] = override_array
+
+        num_items = randint(self.min_items, self.max_items)
```
chrisrcoles, line 106: how will this handle payload attributes, like the `MTOAgents`, that have a `maxItems` attribute?

utils/constants.py
```diff
@@ -14,6 +14,8 @@
 ARRAY_MIN = 1
 ARRAY_MAX = 5
 
+ZERO_UUID = "00000000-0000-0000-0000-000000000000"
```
chrisrcoles, line 17: why do we need zero/ nil uuid?

utils/fields.py
```diff
@@ -83,9 +89,22 @@ def generate_fake_data(self, faker, **kwargs):
         :param kwargs:
         :return: list
         """
-        num_items = randint(self.min_items, self.max_items)
         fake_data = []
 
+        if (overrides := kwargs.get("overrides")) and self.name in overrides:
```
chrisrcoles, line 94: nice syntactic sugar! I've never seen/used this before but am glad you're making use of it in the Python3.8 update!

chrisrcoles: pretty trippy! I thought we had some in golang here for a moment :)

utils/fields.py
sandy-wright: The parser will grab the `minItems` and `maxItems` values from the YAML file and set them as `self.min_items` and `self.max_items` here. That way the `num_items` will always be a random number between those edges. So for MTO Agents, it would always either be 1 or 2 (since we test with all fields as required)

utils/constants.py
sandy-wright: good question! Because we now validate that the ID fields are empty in some endpoints, when we pass it in, it either needs to be a blank but valid UUID field (so the zero UUID), or it needs to be omitted from the payload completely. We unfortunately can't mark these fields as `readOnly` just yet because the same definition is used for the update endpoints, and those endpoints require the ID. It's also a huge pain to try and pop these fields from the dictionary when they're in nested objects as well. 

So the `ZERO_UUID` value ends up being the easiest solution (and this is the value that gets set for a nil UUID in `payload_to_model` so the end result is the same either way).

utils/fields.py
sandy-wright: right?? I really wanted the excuse to use this syntax, so I was very happy when I found the chance haha!

PR #35, opened 2020-12-16
MB-5819-update-mto-agent

NamibiaLT:
## Description
Add an endpoint to update the mto agent.

## Setup

1. run Locust with `make load_test_prime`
2. In the locust load test, select 5 for number of "users to simulate" and for "swarm rate".
3. Check to see how many failures there are.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5819) for this change.

### DIFF
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -264,28 +264,18 @@ class PrimeTasks(PrimeDataTaskMixin, ParserTaskMixin, CertTaskMixin, TaskSet):
     @tag(PrimeObjects.MTO_AGENT.value, "updateMTOAgent")
     @task
     def update_mto_agent(self):
-        mto_agent = self.get_random_data(PrimeObjects.MTO_AGENT)
-        if not mto_agent:
+        mto_shipment = self.get_random_data(
+            PrimeObjects.MTO_SHIPMENT
+        )  # this grabs response payload from the PrimeObjects.MTO_AGENT list when load testing runs
+        if not mto_shipment:
             return  # can't run this task
 
-        payload = self.fake_request("/mto-agent/{mtoAgentID}", "put")
-
-        # These fields need more complicated logic to handle, so remove them for the time being:
-        fields_to_remove = [
-            # "agents", would we need to remove this one if we are trying to update it?
-            "pickupAddress",
-            "destinationAddress",
-            "secondaryPickupAddress",
-            "secondaryDeliveryAddress",
-            "primeEstimatedWeight",
-        ]
-        for f in fields_to_remove:
-            payload.pop(f, None)
-
+        payload = self.fake_request("/mto-shipments/{mtoShipmentID}/agents/{agentID}", "put")
+        mto_agent = mto_shipment["agents"][0]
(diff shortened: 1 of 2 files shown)


tasks/prime.py
```diff
@@ -260,6 +261,31 @@ def update_mto_shipment_address(self):
...
+            PrimeObjects.MTO_SHIPMENT
+        )  # this grabs response payload from the PrimeObjects.MTO_AGENT list when load testing runs
+        if not mto_shipment:
+            return  # can't run this task
+
+        payload = self.fake_request("/mto-shipments/{mtoShipmentID}/agents/{agentID}", "put")
+        mto_agent = mto_shipment["agents"][0]
+        headers = {"content-type": "application/json", "If-Match": mto_agent["eTag"]}
+        resp = self.client.put(
+            prime_path(f"mto-shipments/{mto_shipment['id']}/agents/{mto_agent['id']}"),
```
sandy-wright, line 277: This path also needs to have a leading slash to work: `/mto-shipments/{mto_shipment['id']}/agents/{mto_agent['id']}`

tasks/prime.py
```diff
@@ -260,6 +261,31 @@ def update_mto_shipment_address(self):
...
+            name=prime_path("mto-shipments/:mtoShipmentID/agents/mtoAgentID"),
```
sandy-wright, line 278: After talking with Abigail and Shimona, we decided to use { } curly braces to indicate the IDs in these names - so this line should look exactly like what you passed into the `self.fake_request` function on line 273

tasks/prime.py
```diff
@@ -260,6 +261,31 @@ def update_mto_shipment_address(self):
...
+            data=json.dumps(payload),
+            headers=headers,
+            **self.user.cert_kwargs,
+        )
+
+        resp, success = check_response(resp, "Update MTO Agent", payload)
+
+        if success:
+            self.replace_prime_data(PrimeObjects.MTO_AGENT, mto_agent, resp)
```
sandy-wright, line 287: After pondering this, I think we should be setting a new MTO_SHIPMENT and not an agent here. The reason is because we grab a shipment to start with anyway, so we're never going to use these agents. I'd suggest doing what Chris did for `update_mto_shipment_address`:

```python
if success:
    new_shipment = deepcopy(mto_shipment)
    new_shipment["agents"][0] = resp
    self.replace_prime_data(PrimeObjects.MTO_SHIPMENT, mto_shipment, new_shipment)
```

Let me know if you have any questions about this and we can talk about it in more detail!

sandy-wright (changes requested):
This looks really good, thanks for working on this, Namibia! There are a couple of little issues, the biggest one is the leading slash in the path for the `self.client.put` function call. Once I fixed that locally, the endpoint worked like a charm though! A few edits and it'll be good to go 👍

tasks/prime.py
NamibiaLT: Nice! Thanks for the feedback. I see a HUGE improvement and no failures so far now when running the tests.

sandy-wright (approved):
Looks good, thanks for making those changes!

PR #36, opened 2020-12-16
[MB-3289] Adds task for updateMTOServiceItemStatus handler.

chrisrcoles:
## Description

This PR adds `updateMTOServiceItemStatus` handler to the `SupportTasksSet` for load testing integration. 

## Reviewer Notes

See inline comments. 

## Setup

1. Start and run `mymove` Prime API.
```sh
mylaptop ~ %: cd /mymove
mylaptop ~ %: make server_run
```

2. Load end to end data depended for load test. 
```sh
mylaptop mymove  %: make db_dev_e2e_populate
```

3. Start Locust load testing framework and run `updateMTOServiceItemStatus` test. 
```sh
mylaptop ~ %: cd /milmove_load_testing
mylaptop milmove_load_testing  %: locust -f locustfiles/prime.py --host local --headless -u 5 -r 5
```

4. After starting load tests, ensure handler responds with `200` status code, specifically for `PATCH /prime/v1/mto-shipments/:mtoShipmentID/status`.

* [x] Request review from a member of a different team.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3289) for this change.

### DIFF
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -100,7 +100,7 @@ class PrimeTasks(PrimeDataTaskMixin, ParserTaskMixin, CertTaskMixin, TaskSet):
             # default for local testing
             mto_shipment = {
                 "id": "475579d5-aaa4-4755-8c43-c510381ff9b5",
-                "moveTaskOrderID": "5d4b25bb-eb04-4c03-9a81-ee0398cb779e",
+                "moveTaskOrderID": "99783f4d-ee83-4fc9-8e0c-d32496bef32b",
             }
 
         overrides = {
@@ -296,6 +296,7 @@ class SupportTasks(PrimeDataTaskMixin, ParserTaskMixin, CertTaskMixin, TaskSet):
                 "status": "SUBMITTED",
                 "issueDate": "2020-01-01",
             },
+            "status": "SUBMITTED",
         }
 
         headers = {"content-type": "application/json"}
@@ -321,3 +322,34 @@ class SupportTasks(PrimeDataTaskMixin, ParserTaskMixin, CertTaskMixin, TaskSet):
 
         if success:
             self.set_prime_data(PrimeObjects.MOVE_TASK_ORDER, mto_data)
+
+    @tag(PrimeObjects.MTO_SERVICE_ITEM.value, "updateMTOServiceItemStatus")
+    @task
+    def update_mto_service_item_status(self):
+        mto_shipment = self.get_random_data(PrimeObjects.MTO_SHIPMENT)
+        # if we don't have an mto shipment we can't run this task
+        if not mto_shipment:
+            return
+
+        # if we don't have any service items created for mto shipment can't run this task
+        if mto_shipment.get("mtoServiceItems") is None:
+            return
+
+        payload = self.fake_request("/service-items/{mtoServiceItemID}/status", "patch")
+        mto_service_item = mto_shipment["mtoServiceItems"][0]
+        headers = {"content-type": "application/json", "If-Match": mto_service_item["eTag"]}
+
+        resp = self.client.patch(
+            support_path(f"/service-items/{mto_service_item['id']}/status"),
+            name=support_path("/mto-shipments/:mtoShipmentID/status"),
+            data=json.dumps(payload),
+            headers=headers,
+            **self.user.cert_kwargs,
+        )
+
+        mto_service_item_data, success = check_response(resp, "Update MTO service item status")
+
+        if success:
+            new_shipment = deepcopy(mto_shipment)
+            new_shipment["mtoServiceItems"][0] = mto_service_item_data
+            self.replace_prime_data(PrimeObjects.MTO_SHIPMENT, mto_shipment, new_shipment)
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -361,7 +361,7 @@ class SupportTasks(PrimeDataTaskMixin, ParserTaskMixin, CertTaskMixin, TaskSet):
 
         resp = self.client.patch(
             support_path(f"/service-items/{mto_service_item['id']}/status"),
-            name=support_path("/service-items/:mtoShipmentID/status"),
+            name=support_path("/service-items/{mtoShipmentID}/status"),
             data=json.dumps(payload),
             headers=headers,
             **self.user.cert_kwargs,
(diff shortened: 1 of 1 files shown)


tasks/prime.py
```diff
@@ -347,3 +347,34 @@ def create_move_task_order(self):
...
+    @task
+    def update_mto_service_item_status(self):
+        mto_shipment = self.get_random_data(PrimeObjects.MTO_SHIPMENT)
+        # if we don't have an mto shipment we can't run this task
+        if not mto_shipment:
+            return
+
+        # if we don't have any service items created for mto shipment can't run this task
+        if mto_shipment.get("mtoServiceItems") is None:
+            return
```
sandy-wright, line 361: Is there a reason you aren't grabbing a MTO_SERVICE_ITEM to start with? I don't think there's a reason to have the shipment around too. It seems like the issues with `create_mto_service_items` have been mostly fixed - or maybe I've been lucky with successful tests that haven't used the bad options haha

tasks/prime.py
```diff
@@ -347,3 +347,34 @@ def create_move_task_order(self):
...
+
+        payload = self.fake_request("/service-items/{mtoServiceItemID}/status", "patch")
+        mto_service_item = mto_shipment["mtoServiceItems"][0]
+        headers = {"content-type": "application/json", "If-Match": mto_service_item["eTag"]}
+
+        resp = self.client.patch(
+            support_path(f"/service-items/{mto_service_item['id']}/status"),
+            name=support_path("/mto-shipments/:mtoShipmentID/status"),
```
sandy-wright, line 369: looks like a copy paste error here ☺️ - should be "/service-tems/{mtoServiceItemID}/status" < we're going to go back at make the curly braces standard here since that's what it looks like in the API YAML file

tasks/prime.py
```diff
@@ -100,7 +100,7 @@ def create_mto_service_item(self):
             # default for local testing
             mto_shipment = {
                 "id": "475579d5-aaa4-4755-8c43-c510381ff9b5",
-                "moveTaskOrderID": "5d4b25bb-eb04-4c03-9a81-ee0398cb779e",
+                "moveTaskOrderID": "99783f4d-ee83-4fc9-8e0c-d32496bef32b",
```
sandy-wright, line 103: thanks for catching this!!

sandy-wright (commented):
Looks good! I left a comment about maybe simplifying the function to grab a MTO_SERVICE_ITEM to start with so you don't have to go through a shipment - that should make things a little clearer. There's also a tiny copy paste error that would be good to change. Tests all ran really well though!

tasks/prime.py
chrisrcoles: Hey thanks for this comment. I had more success grabbing the `mto_shipment` first. Originally, when I implemented this, it look more like:
```
    mto_service_item = self.get_random_data(PrimeObjects.MTO_SERVICE_ITEM)
        # if we don't have an mto shipment we can't run this task
        if not mto_service_item:
            return

        payload = self.fake_request("/service-items/{mtoServiceItemID}/status", "patch")
        headers = {"content-type": "application/json", "If-Match": mto_service_item["eTag"]}

        resp = self.client.patch(
            support_path(f"/service-items/{mto_service_item['id']}/status"),
            name=support_path("/service-items/:mtoShipmentID/status"),
            data=json.dumps(payload),
            headers=headers,
            **self.user.cert_kwargs,
        )

        mto_service_item_data, success = check_response(resp, "Update MTO service item status")

        if success:
            self.replace_prime_data(PrimeObjects.MTO_SHIPMENT, mto_service_item, mto_service_item_data)
```
As you can see from above, this grabs the `mto_service_item` initially. The problem with this implementation is that it results in more `409` status codes. When implementing this load test as you suggest, I run into more failures. On average, 20-40% of the calls fail whereas with the current implementation as I've proposed we have 100% success rate. I'd rather keep it the way it is given that we have a higher success rate. Please let me know if there's something I could be doing to get a higher success rate. See below: 
Screen Shot 2020-12-16 at 4 55 56 PM

tasks/prime.py
chrisrcoles: I've updated this! Thanks for catching this!

tasks/prime.py
chrisrcoles: Take a look at the success rates. 
Screen Shot 2020-12-16 at 5 04 00 PM

sandy-wright: hmm, I'm not sure if this is a copy paste error now, but at the bottom you're replacing the info for a MTO_SHIPMENT, not a MTO_SERVICE_ITEM. This means the old service items with the old eTags would still be in the list. Is it possible that was causing your problem?

sandy-wright: I also think some of these errors 409 are expected, and there's nothing preventing this specific error from happening by grabbing a service item from a shipment (a shipment can also have an APPROVED service item that the endpoint tries to APPROVE again). I suspect that if you ran both versions for a sufficiently long period of time, the rate of success would even out. There's the random element at play too

chrisrcoles: Ok! Great! I've updated this to grab the `mto_service_item` instead of the `mto_shipment` and updated the `replace_prime_data` call to replace the `MTO_SERVICE_ITEM` instead of the `MTO_SHIPMENT`.

tasks/prime.py
```diff
@@ -347,3 +347,27 @@ def create_move_task_order(self):
...
+        # if we don't have an mto shipment we can't run this task
+        if not mto_service_item:
+            return
+
+        payload = self.fake_request("/service-items/{mtoServiceItemID}/status", "patch")
+        headers = {"content-type": "application/json", "If-Match": mto_service_item["eTag"]}
+
+        resp = self.client.patch(
+            support_path(f"/service-items/{mto_service_item['id']}/status"),
+            name=support_path("/service-items/:mtoShipmentID/status"),
```
sandy-wright, line 364: Looks like the ID part here still needs to be changed to `{mtoServiceItemID}` instead of `:mtoShipmentID`

sandy-wright (approved):
LGTM! There's still a tiny typo, but other than that it's good to go ✨

tasks/prime.py
chrisrcoles: Thanks for this! I've updated it!

PR #37, opened 2020-12-18
MB-5820 Load testing UpdateMTOServiceItem

shimonacarvalho:
## Description

This ticket was to add a task for the endpoint updateMTOServiceItem. This endpoint needs DDDSIT or DOPSIT items to update. Those items are created when you create a DOFSIT or DDFSIT item as those create a trio of SIT items. 

So this code adds tasks to request DOFSIT or DDFSIT creation, and to also request updating the DDDSIT or DOPSIT items. 

Some fixes that went in this code
- missing agents check
- log url and method when we have an error response
- make createMTOServiceItems accept and apply overrides
- update response strings to use the operationID for consistency

## Setup
You need a lot of requests for there to be enough to create the specific types you want. 

- In the locust load test, select 50/10 for number of "users to simulate" and for "swarm rate".
- Check to see how many failures there are.

Then watch the Locust statistics page and wait for at least a few hits on the `/prime/v1/mto-service-items/:mtoServiceItemID` patch command. 

Then check your logs and see if you can find some successful updateMTOServiceItem calls. Note there is a different endpoint called updateMTOServiceItemStatus. 

You may see the following log which is okay.
```
update_mto_service_item recvd mtoServiceItem from store. Discarding because reServiceCode not in [DDDSIT, DOPSIT]
```

* [ ] Did you catch some successful hits to the endpoint?
* [ ] Do the failures make sense? 
* [ ] 500s are no good, if you catch them, add a ticket against milmove with the logs from the server. 

## References

* [MB-5820 Prime API: Load Testing - UpdateMTOServiceItem](https://dp3.atlassian.net/browse/MB-5820)

### DIFF
--- a/tasks/base.py
+++ b/tasks/base.py
@@ -33,9 +33,11 @@ def check_response(response, task_name="Task", request=None):
         logger.error(f"⚠️\n{json.dumps(json_response, indent=4)}")
         if request:
             try:
-                logger.error(f"Request data:\n{json.dumps(request, indent=4)}")
+                logger.error(
+                    f"Request data:\n{response.request.method} {response.request.url}\n{json.dumps(request, indent=4)}"
+                )
             except (json.JSONDecodeError, TypeError):
-                logger.error(f"Request data:\n{request}")
+                logger.error(f"Request data:\n{response.request.method} {response.request.url}\n{request}")
 
         return json_response, False
 
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -300,7 +300,7 @@ class PrimeTasks(PrimeDataTaskMixin, ParserTaskMixin, CertTaskMixin, TaskSet):
             logger.error("⛔️ update_mto_service_item recvd mtoServiceItem without reServiceCode \n{mto_service_item}")
             return
 
-        if re_service_code != "DDDSIT" and re_service_code != "DOPSIT":
+        if re_service_code not in ["DDDSIT", "DOPSIT"]:
             logging.info(
                 "update_mto_service_item recvd mtoServiceItem from store. Discarding because reServiceCode not in [DDDSIT, DOPSIT]"
             )
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -88,7 +88,7 @@ class PrimeTasks(PrimeDataTaskMixin, ParserTaskMixin, CertTaskMixin, TaskSet):
     @task
     def fetch_mto_updates(self):
         resp = self.client.get(prime_path("/move-task-orders"), **self.user.cert_kwargs)
-        check_response(resp, "Fetch MTOs")
+        check_response(resp, "fetchMTOUpdates")
 
     @tag(PrimeObjects.MTO_SERVICE_ITEM.value, "createMTOServiceItem")
     @task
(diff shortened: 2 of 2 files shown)


tasks/base.py
```diff
@@ -33,9 +33,11 @@ def check_response(response, task_name="Task", request=None):
         logger.error(f"⚠️\n{json.dumps(json_response, indent=4)}")
         if request:
             try:
-                logger.error(f"Request data:\n{json.dumps(request, indent=4)}")
+                logger.error(
+                    f"Request data:\n{response.request.method} {response.request.url}\n{json.dumps(request, indent=4)}"
+                )
             except (json.JSONDecodeError, TypeError):
-                logger.error(f"Request data:\n{request}")
+                logger.error(f"Request data:\n{response.request.method} {response.request.url}\n{request}")
```
sandy-wright, line 40: This is such a nice addition

tasks/prime.py
```diff
@@ -88,11 +88,11 @@ class PrimeTasks(PrimeDataTaskMixin, ParserTaskMixin, CertTaskMixin, TaskSet):
     @task
     def fetch_mto_updates(self):
         resp = self.client.get(prime_path("/move-task-orders"), **self.user.cert_kwargs)
-        check_response(resp, "Fetch MTOs")
+        check_response(resp, "fetchMTOUpdates")
```
sandy-wright, line 91: thanks for changing these! 🙌

tasks/prime.py
```diff
@@ -238,6 +266,8 @@ def update_mto_agent(self):
         )  # this grabs response payload from the PrimeObjects.MTO_AGENT list when load testing runs
         if not mto_shipment:
             return  # can't run this task
+        if mto_shipment.get("agents") is None:
```
sandy-wright, line 269: Thank you for handling this error 😄 ✨

tasks/prime.py
```diff
@@ -250,13 +280,49 @@ def update_mto_agent(self):
...
+        if not mto_service_item:
+            return  # can't run this task
+
+        try:
+            re_service_code = mto_service_item["reServiceCode"]
+        except KeyError:
+            logger.error("⛔️ update_mto_service_item recvd mtoServiceItem without reServiceCode \n{mto_service_item}")
+            return
+
+        if re_service_code != "DDDSIT" and re_service_code != "DOPSIT":
```
sandy-wright, line 303: Another pythonic way to write this would be:

```python
if re_service_code not in ["DDDSIT", "DOPSIT"]:
```

By no means required, it's just a little more concise

sandy-wright (approved):
This looks great! It does take a while to get to the departure SIT service items for an update because we're grabbing a random service item from the stack, and it goes through a lot that aren't eligible for the update.

As a future enhancement, we should consider adding a separate list to the local store just for departure SIT items that the `create_mto_service_items` endpoint populates when it gets a DDD/DOPSIT item. Then we could update this task to pull directly from that list. I can add a ticket to the backlog for that!

tasks/prime.py
shimonacarvalho: ooh nice! And matches the error message.

PR #38, opened 2020-12-18
Adds `f` so error message will print mtoServiceItem data

abbyoung:
## Description

One-character change so our `mtoServiceItem` data will log out with the error message.


sandy-wright (approved):
LGTM! Ran load tests and confirmed that we see the output of the service item now! Also luck was on my side and it started hitting this endpoint right away this time, whew

PR #42, opened 2020-12-29
MB-3240 Added updatePaymentRequestStatus task to load testing

shimonacarvalho:
## Description

Added a task for `updatePaymentRequestStatus` to load testing, checked that there were some successful runs. 

## Setup

Please run load testing with 
```sh
make load_test_prime
```

You need a lot of requests for there to be enough to create the specific types you want.

- In the locust load test, select 50/10 for number of "users to simulate" and for "swarm rate".
- Check to see how many failures there are.

Then watch the Locust statistics page and wait for at least a few hits on the /support/v1/payment-requests/{paymentRequestID}/status patch command.

### What to check for
Then check your logs and see if you can find some successful updatePaymentRequestStatus calls. 
```
tasks.base: ℹ️ updatePaymentRequestStatus status code: 200
tasks.base: ℹ️ updatePaymentRequestStatus successfully completed!
```

- [ ] Did you catch some successful hits to the endpoint?
- [ ] Do the failures make sense?
- [ ] 500s are no good, if you catch them, add a ticket against milmove with the logs from the server.
Note, I ticketed one already. 500 error due to this reason has been ticketed.
`All PaymentServiceItems must be approved or denied to review this PaymentRequest`

## References

* [MB-3240 Support API: Load Testing - add task for updatePaymentRequestStatus](https://dp3.atlassian.net/browse/MB-3240)

### DIFF
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -410,6 +410,7 @@ class SupportTasks(PrimeDataTaskMixin, ParserTaskMixin, CertTaskMixin, TaskSet):
                     "agency": "MARINES",
                     "email": "swinglehurst@example.com",
                     "rank": "E_3",
+                    "dodID": "4586736251",
                 },
                 "entitlement": {"nonTemporaryStorage": False, "totalDependents": 47},
                 "orderNumber": "32",
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -136,30 +136,6 @@ class PrimeTasks(PrimeDataTaskMixin, ParserTaskMixin, CertTaskMixin, TaskSet):
         if success:
             self.set_prime_data(PrimeObjects.MTO_SERVICE_ITEM, resp)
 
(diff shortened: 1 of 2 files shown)


tasks/prime.py
```diff
@@ -409,17 +409,19 @@ def create_move_task_order(self):
                     "lastName": "Swinglehurst-Walters",
                     "agency": "MARINES",
                     "email": "swinglehurst@example.com",
+                    "rank": "E_3",
```
sandy-wright, line 412: It looks like we need to set a `"dodID"` in the Customer as well - I added it in the branch to test.

sandy-wright (approved):
LGTM! I had to add a field to the create MTO payload to make it work initially, but after that, it worked like a charm ✨ . I think the field issue will be resolved when you handle the merge conflicts with master, so it should be good to go

tasks/prime.py
shimonacarvalho: Okay thanks!

PR #44, opened 2020-12-29
MB-3293 Changed createMoveTaskOrder task in loadtesting to use fake data

shimonacarvalho:
## Description

CreateMoveTaskOrder is an endpoint. The loadtesting task for this endpoint was using a fixed, hardcoded set of data, which is not great for really exercising the system. 
This PR replaces that with fake data. 

Note - there are a number of things that can't be totally random and so they are cleaned up in the custom validation function in the supportAPIParser. Please make sure they make sense to you. 

## Setup

Please run load testing with 
```sh
make load_test_prime
```
- In the locust load test, select 5/10 for number of "users to simulate" and for "swarm rate".

Then watch the Locust statistics page and wait for at least a few hits on the /support/v1/move-task-orders/ POST command.

### What to check for
Then check your logs and see if you can find successful createMoveTaskOrder calls. 
```
tasks.base: ℹ️ createMoveTaskOrder status code: 201
tasks.base: ℹ️ createMoveTaskOrder successfully completed!
```

In the  UI you should see success for the following endpoint
` POST /support/v1/move-task-orders     `

- [ ] Were the hits to the endpoint mostly successful? 
- [ ] Do the 400 failures make sense to you?
- [ ] 500s are not ok. If you catch them, add a ticket against milmove with the logs from the server.

## References

* [MB-3293 Load Testing: Implement fake data tool with create_move_task_order - Jira](https://dp3.atlassian.net/browse/MB-3293)
* [Getting Started Guide to Load Testing for MilMove](https://github.com/transcom/milmove_load_testing/#getting-started)


utils/parsers.py
```diff
@@ -425,3 +425,33 @@ class SupportAPIParser(PrimeAPIParser):
...
+        """
+        Custom post-data generation validation for the Support API. Note that request_data is mutable and directly
+        modified, and therefore doesn't need to be returned.
+        """
+        if path == "/move-task-orders" and method == "post":
+
+            move_task_order = request_data
+            move_orders = request_data["moveOrder"]
+            customer = move_orders["customer"]
+            entitlement = move_orders["entitlement"]
```
sandy-wright, line 439: Splitting it out this way makes it really clear later on! I like it. There are so many nested objects in this payload, oof

tasks/prime.py
```diff
@@ -374,31 +374,18 @@ def update_mto_shipment_status(self):
...
-                "uploadedOrdersID": "c26421b0-e4c3-446b-88f3-493bb25c1756",
-                "ordersType": "GHC",
-                "reportByDate": "2020-01-01",
-                "status": "SUBMITTED",
-                "issueDate": "2020-01-01",
-            },
-            "status": "SUBMITTED",
+            # Currently the state engine will fail on creating all mto service items
+            # unless the state is APPROVED
+            "status": "APPROVED",
```
sandy-wright, line 386: Was the issue you ran into that this status was applied to both the `move` and the `moveOrder`? I think I re-understand the way it works now (it's been a while since I wrote this 😅) - you can nest overrides like you do in a payload. So the above could also be written as:

```python
overrides = {
        "status": "APPROVED",
        "moveOrder": {
            "status": "APPROVED",
            "destinationDutyStationID": "71b2cafd-7396-4265-8225-ff82be863e01",
            "originDutyStationID": "1347d7f3-2f9a-44df-b3a5-63941dd55b34",
            "uploadedOrdersID": "c26421b0-e4c3-446b-88f3-493bb25c1756",
        }
}
```

The upside being that that is more explicit about what's being overridden, the downside being that you can't shortcut the setting of "status" anymore. You have to explicitly set that override for the "moveOrder" too. 

So these are the two ways overrides can be used - set ANY field with that name to a certain value, or explicitly set the values in nested objects. I think if you did something like `"moveOrder": {}` it would actually generate fake data for this object without repeating any of the overrides from the top level (but I didn't test that to be sure). Man, I sure overdid it with this hahaha

sandy-wright: Also, I believe you don't need to set the `moveTaskOrder` "status" field to APPROVED - the `makeAvailableToPrime` endpoint should also approve the MTO. But either way is fine with me.

sandy-wright (approved):
LGTM! I'm happy that you used the custom hook functions I put in the parser, and I'm glad I got a chance to reevaluate the strange code I wrote for overrides 😄

utils/parsers.py
shimonacarvalho: Yes, a lot. It's a tough one.

tasks/prime.py
shimonacarvalho: Oh I see, I need to nest it. I guess what threw me off was that it DID show up in the moveOrder but it also showed up in every other object. Like just, everything hehe. Let me see if I can get it to work exactly right.

PR #45, opened 2020-12-31
MB-6088-unit-tests-check-response-func

NamibiaLT:
## Description

Write unit tests for the check response function in `tasks/base.py`

## Setup

To run the tests use this command

```python
python3 -m pytest -v
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6088) for this change.


tasks/tests/test_base.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+from tasks.base import check_response
+from requests import Response
```
sandy-wright, line 3: Same comment as above, it would be best to switch the order of these imports and put a blank line in between. Also, I like to put a docstring at the top of the file to briefly explain what it's for, like:

```
# -*- coding: utf-8 -*-
""" This file tests the classes and functions in tasks/base.py """
# imports after that
``` 

This is a "Sandy standard", not a community standard, so you can feel free to ignore if you like! 😄

tasks/tests/test_base.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def test_check_response_happy_path():
+    response = Response()
+    response.status_code = 200
+    response._content = '{"field": "value"}'
+
+    # Happy Path:
+    assert check_response(response) == ({"field": "value"}, True)
```
sandy-wright, line 12: I like to put docstrings under most functions if possible, so maybe a short sentence that describes what's being tested would be nice?

```python
def test_check_response_happy_path():
    """
    This tests that the check_response function will return the response content and a True value 
    when it receives a Response object with a 2xx status code and valid JSON in the content.
    """
```

This is also SUPER optional, but can be nice to have later on.

sandy-wright (approved):
LGTM, and yay tests!! I left comments about stylistic things that could be tightened up, but it's good to go otherwise.

tasks/tests/test_base.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+""" This file tests the classes and functions in tasks/base.py """
+
+from requests import Response
+from tasks.base import check_response
```
sandy-wright, line 5: almost perfect! There would usually be an empty line in between these two import groups

tasks/base.py
```diff
@@ -1,24 +1,26 @@
...
 """ tasks/base.py is for code used internally within the tasks package. """
+from utils.base import ImplementationError
+
 import logging
 import json
+from requests import Response
 
 from locust import TaskSet
 
-from utils.base import ImplementationError
```
sandy-wright, line 8: The `utils.base` import should still be at the very bottom, and `requests` should be grouped with `locust` (it's not a Python-base package, you have to install it)

tasks/tests/test_base.py
```diff
@@ -0,0 +1,56 @@
...
+    bad_status_code._content = '{"this_is": "ok"}'
+
+    mocker.patch.object(logger, "error")
+    assert check_response(bad_status_code, "Task", '{ "a":1 }') == ({"this_is": "ok"}, False)
+    assert logger.error.call_count == 2
+    expected_calls = [
+        mocker.call('⚠️\n{\n    "this_is": "ok"\n}'),
+        mocker.call('Request data:\n"{ \\"a\\":1 }"'),
+    ]
+    assert logger.error.call_args_list == expected_calls
```
sandy-wright, line 56: This mock logger is super cool! Thanks for figuring this out and adding it 😄

PR #50, opened 2021-01-15
[MB-3077] Prime workflow setup for load testing

shimonacarvalho:
## Description

This PR creates a prime workflow infrastructure for us to put in entire sequential workflows. It covers 4 endpoints in the workflow:

- createMoveTaskOrder
- makeMoveTaskOrderAvailable
- updateMTOPostCounselingInformation
- fetchMTOUpdates

## Setup

- run Locust with make `load_test_prime_workflow`
- In the locust load test, select 25/10 for number of "users to simulate" and for "swarm rate".
- Check that all four endpoints are hit
- Check that there are ZERO failures. 

Unlike the discrete task based load testing, the prime_workflow taskset will **only test happy paths**. Therefore we expect no failures.
You should see hits to the above endpoints. You should see ZERO failures.

There are also tests setup for the storage aspect of prime workflow
You can run them with 
`pytest tasks/tests/test_prime_workflow.py` 

Should also have ZERO failures.

## References

[MB-3077 Load Testing: Set up a locustfile that for the Prime Workflow - Jira](https://dp3.atlassian.net/browse/MB-3077)


tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+        """Return object of type requested. If no id is provided return a random element.
+        If id is provided, return the specific element or raise error
+        """
+
+        # Get the nested array from the current move, or None if it doesn't exist
+        array_key = self._get_array_name(object_key)
+        nested_array = self.current_move.get(array_key, None)
+
+        # If nested array does not exist or is empty, return None
+        if nested_array is None or nested_array == []:
```
sandy-wright, line 190: Same as above, can be: `if not nested_array:`

tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+        if id:
+            for elem in nested_array:
+                if elem["id"] == id:
+                    requested_object = elem
+                    break
+        # If id was not provided select an object at random
+        else:
+            requested_object = random.choice(nested_array)
+
+        return requested_object
```
sandy-wright, line 204: I would simplify this to:

```python
# If id was provided, return the specific object:
if id:
    for elem in nested_array:
        if elem["id"] == id:
            return elem

    return None

# If id was not provided select an object at random:
return random.choice(nested_array)
```

tasks/tests/test_prime_workflow.py
```diff
@@ -0,0 +1,218 @@
...
+
+        self.locust = TestWorkflowUser(self.environment)
+        self.taskset = WorkflowTasks(self.locust)
+
+    def test_add_stored_mto(self):
+        # Under test: add_stored function
+        # Mocked: Data objects are simplified versions
+
+        # Scenario: We add a move, and then another move
+        # Expectation: It should store the move into current move and then replace it, since we only keep one move.
```
sandy-wright, line 30: These comments should be in the docstring for this function, but that's a really minor nitpick.

tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+    There are multiple workflows for different scenarios such as a single shipment HHG move, or an HHG move with a SIT.
+    Multiple instances of the tasks will run during load testing and will interleave their hits to the server.
+    However they will run in sequence and should all complete without any errors.
+    """
+
+    # current_move is the move the whole task will be acting on
+    # all updates to the move and nested objects will be stored to this mto
+    current_move = None
+    # workflow_descriptor
+    workflow_descriptor = None
```
sandy-wright, line 32: Is this ever used?

tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+
+    # WORKFLOWS
+    @tag("hhgMove")
+    @task
+    def hhg_move(self):
+        """
+        This is a basic HHG move workflow where we have one shipment and no additional requested service items.
+        To create a new flow, copy this function, but don't forget to add the workflow to the tasks array in the on_start function
+        """
+        self.workflow_title = "HHG Move Basic"
```
sandy-wright, line 50: I don't think this is ever used more than once, so I would delete this variable and just hardcode it into the log statement later on.

sandy-wright (commented):
Thank you for working on this! This was a LOT of work, and you implemented a bunch of great solutions and improvements. Plus it works like a charm, and the tests are 😚👌✨  

I left a lot of comments - let me know if you have any questions or would like to go over them together. I was pretty picky/thorough, so this was really everything I could see. Some of them are negotiable (I can't say which ones because I'm not sure what you could convince me of yet haha).

One other suggestion would be to rename the `locustfiles/workflow.py` file to `locustfiles/prime_workflow.py`, since this app will likely be extended to include workflows for other parts of the app. It's pretty common in python to have files in different packages with the same name, but if you want to make it different from the `tasks/` prime files, you could go with `locustfiles/prime_users.py` and `locustfiles/prime_workflow_users.py` instead.

tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+    MTO_SERVICE_ITEM,
+    PAYMENT_REQUEST,
+)
+from utils.base import ImplementationError
+
+logger = logging.getLogger(__name__)
+
+
+@tag("workflow")
+class WorkflowTasks(PrimeTasks, SupportTasks):
```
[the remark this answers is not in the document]
shimonacarvalho: Okay, I can do that. My thinking was each workflow is going to have actions from prime and office and customer potentially, so I was having trouble separating them. This one does have actions that would be taken by TOO/TIO for e.g. The main difference I guess this one is api-only, whereas the others will be app+api but those aren't great names. 
But in the end I couldn't think of a good way to separate this workflow from the other one so I'll do PrimeWorkflowTasks

tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+        # Move steps for this workflow
+        super().create_move_task_order()
+        super().update_post_counseling_information()
+        logger.info(f"{self.workflow_title} - Created move and completed counseling {self.current_move['id'][:8]}")
+
+    # STORAGE FUNCTIONALITY
+    """ We only store one current move in the sequential workflow and keep it updated as we move through the workflow.
+        These functions override functions in the PrimeDataTaskMixin and will get automatically called by the tasks
+        in PrimeTasks and SupportTasks.
+    """
```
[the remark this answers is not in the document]
shimonacarvalho: Oh good to know!

tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+    def add_stored(self, object_key, object_data):
+        """Adds the object to the main move."""
+
+        # If it's a moveTaskOrder, we only keep track of one move per workflow, so add actually replaces.
+        if object_key == MOVE_TASK_ORDER:
+            self.current_move = object_data
+
+        # If it's the following object types, we add them into nested array in the MTO
+        elif object_key in [MTO_SHIPMENT, MTO_SERVICE_ITEM, PAYMENT_REQUEST]:
+            self._add_to_mto_nested(object_key, object_data)
```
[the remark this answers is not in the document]
shimonacarvalho: I saw MTO_AGENT as one of the object types, and that one cannot be used in this way because it is not directly nested into an array in the mto.

tasks/prime_workflow.py
shimonacarvalho: Originally I had subfunctions with bits of the flow, and this was a way for each subfunction to log meaningful messages about what work they had done, like "HHG Move Basic - Completed creation and counseling". But since my ticket was only for a tiny piece of the flow, I removed the subfunctions. I think it will be useful once we have the whole flow?

tasks/tests/test_prime_workflow.py
shimonacarvalho: Put the Under test and Mocked comments into docstrings. The Scenario and Expectation are regular comments.

tasks/prime_workflow.py
shimonacarvalho: oops, it was supposed to be workflow_title.

tasks/prime_workflow.py
shimonacarvalho: Done

tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def get_stored(self, object_key, id=None):
+        """ Override the get_stored function from PrimeDataTaskMixin """
+
+        # Return whichever part of the move was requested - such as a shipment:
+        if object_key == MOVE_TASK_ORDER:
+            return self.current_move
+
+        if object_key == MTO_SHIPMENT:
+            return self._get_mto_nested(object_key, id)
```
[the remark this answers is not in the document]
shimonacarvalho: See next comment on MTO_AGENT.

tasks/prime_workflow.py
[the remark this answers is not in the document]
shimonacarvalho: Done.

tasks/prime_workflow.py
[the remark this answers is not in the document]
sandy-wright: Ohhh, that is a fair point. I think that MTO_AGENT was added and then never used, so we should probably delete it. If it's still there though, we can leave these conditionals and then update it when we remove it later.

tasks/prime_workflow.py
sandy-wright: That's fair! Part of my cognitive dissonance with this was that it was different from the `workflow_descriptor` on the class level, but it makes more sense when they're the same thing

tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+                )
+            )
+
+        # Get the nested array from the current move, or None if it doesn't exist
+        array_key = self._get_array_name(object_key)
+        nested_array = self.current_move.get(array_key, None)
+
+        # If nested array is not an array create one
+        if nested_array is None:
+            nested_array = []
```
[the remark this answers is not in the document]
sandy-wright: I would still advocate for simplifying this to the one line version

tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+                    f"does not match the current move's id of {self.current_move['id']}."
+                )
+            )
+
+        # Get the nested array from the current move, or None if it doesn't exist
+        array_key = self._get_array_name(object_key)
+        nested_array = self.current_move.get(array_key, None)
+
+        # If nested array does not exist or is empty, raise error, we should have found the object to replace
+        if nested_array is None or nested_array == []:
```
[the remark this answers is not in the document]
sandy-wright: Note that `not nested_array` will catch an empty list - it's not like Go. So you don't need to check for an empty list.

tasks/prime_workflow.py
sandy-wright: I mentioned this above, but `if not nested_array` will catch an empty list, so you don't need to also check for an empty list.

tasks/prime_workflow.py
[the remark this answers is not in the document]
shimonacarvalho: oh sorry I didn't see that somehow! Yours is very nice! Will fix.

tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+                self.current_move = new_data
+            else:
+                raise ImplementationError(
+                    f"Cannot update move with object because its ID of {old_data['id']} does not "
+                    f"match the current move's id of {self.current_move['id']}."
+                )
+        elif object_key in [MTO_SHIPMENT, MTO_SERVICE_ITEM, PAYMENT_REQUEST]:
+            self._update_mto_nested(object_key, old_data, new_data)
+
+    def _add_to_mto_nested(self, object_key, object_data):
```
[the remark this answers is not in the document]
shimonacarvalho: Sure!

tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+        object_array_map = {
+            MTO_SERVICE_ITEM: "mtoServiceItems",
+            MTO_SHIPMENT: "mtoShipments",
+            PAYMENT_REQUEST: "paymentRequests",
+        }
+        try:
+            array_key = object_array_map[object_key]
+        except KeyError:
+            logger.exception(f"An unexpected type of {object_key} was passed into _add_to_mto_nested")
+            return None
```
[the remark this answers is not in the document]
shimonacarvalho: Done

tasks/prime_workflow.py
shimonacarvalho: Clearly my reading capacity is terrible on "monday" mornings.

sandy-wright (approved):
LGTM!

tasks/prime_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+        return array_key
```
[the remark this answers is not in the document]
shimonacarvalho: Done.

tasks/prime_workflow.py
shimonacarvalho: Done.

PR #51, opened 2021-02-09
MB-6656 Dynamically retrieve required move IDs to create MTOs in Experimental

sandy-wright:
## Description

This PR removes the hardcoded IDs for:
 - `contractorID`
 - `uploadedOrdersID`
 - `destinationDutyStationID`
 - `originDutyStationID`

All of these values are required for the Support API endpoint `createMoveTaskOrder`, which is the pivotal request in our load testing process because it gives us MTOs to work with. In order to run these tests in experimental, we must grab these values from MTOs that already exist in the experimental database. We use the Support API `getMoveTaskOrder` endpoint to do this with the moves we get back from the Prime API's `fetchMTOUpdates`.

## Setup

To test this update, you must work with the `mymove` updates in this PR: https://github.com/transcom/mymove/pull/5861. For a local test, checkout and run the server off the branch `sw-upd-create-mto-for-load-testing`. For a test in experimental, you will either need to wait until this PR is merged with master, or claim and deploy this branch to experimental yourself. 

### Local Testing

Start up your local server:
```sh
make db_dev_e2e_populate server_run
```

And run the load tests:
```sh
make load_test_prime
```

or:
```sh
make load_test_prime_workflow
```

If you see successful calls to the `createMoveTaskOrder` endpoint (`POST | /support/v1/move-task-orders`), then the code works! You can peek in the logs and look for a line like `☑️ Set default MTO IDs for createMoveTaskOrder:` to see which IDs were set.

### Testing in Experimental

Build and run the docker container:
```sh
docker-compose build
docker-compose up
```

Navigate to http://localhost:8089/ to use the web UI to run the test. Spawn 2 users and view the results closely. Once the `createMoveTaskOrder` endpoint has been hit a couple of times, **stop the load test. DO NOT let the test run unmonitored in experimental. We DO NOT want to flood this environment with new moves (yet).**

You should see some successfully created moves. Yay!

### Running Unit Tests

To run the unit tests for the new code, you will first need to install the new required package (in your pyenv virtual environment):
```sh
pyenv activate locust-venv
pip install -r requirements-dev.txt
```

Then use `pytest` to run the tests:
```sh
pytest tasks/tests/test_prime.py
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6656) for this change.

### DIFF
--- a/locustfiles/prime.py
+++ b/locustfiles/prime.py
@@ -27,7 +27,6 @@ class PrimeUser(MilMoveHostMixin, HttpUser):
     # These are locust HttpUser attributes that help define and shape the load test:
     wait_time = between(0.25, 9)  # the time period to wait in between tasks (in seconds, accepts decimals and 0)
     tasks = {PrimeTasks: 1}  # the set of tasks to be executed and their relative weight
-    weight = 5
 
 
 class SupportUser(MilMoveHostMixin, HttpUser):
@@ -43,7 +42,6 @@ class SupportUser(MilMoveHostMixin, HttpUser):
 
     wait_time = between(0.25, 9)
     tasks = {SupportTasks: 1}
-    weight = 1
 
 
 @events.test_stop.add_listener
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -42,7 +42,12 @@ class PrimeDataStorageMixin:
 
     DATA_LIST_MAX: int = 50
     # contains the ID values needed when creating moves using createMoveTaskOrder:
-    default_mto_ids: Dict[str, str] = {}
+    default_mto_ids: Dict[str, str] = {
+        "contractorID": "",
+        "destinationDutyStationID": "",
+        "originDutyStationID": "",
+        "uploadedOrdersID": "",
+    }
     local_store: Dict[str, list] = {
         MOVE_TASK_ORDER: [],
         MTO_SHIPMENT: [],
@@ -150,8 +155,9 @@ class PrimeDataStorageMixin:
         :param moves: list of JSON/dict objects
         :return: None
         """
-        if all(self.default_mto_ids.values()):
-            return  # if we already have all of the ID values, we don't need to run through this logic again
+        # Checks that we have a full set of MTO IDs already and halts processing if so:
+        if self.default_mto_ids and all(self.default_mto_ids.values()):
+            return
 
         headers = {"content-type": "application/json"}
         for move in moves:
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -117,10 +117,12 @@ class PrimeDataStorageMixin:
         """
         data_list = self.local_store[object_key]
 
-        try:
-            data_list.remove(old_data)
-        except ValueError:
-            pass  # this is fine, we didn't want this value in the list anymore anyway
+        # Remove all instances of the stored object, in case multiples were added erroneously:
+        while True:
+            try:
+                data_list.remove(old_data)
+            except ValueError:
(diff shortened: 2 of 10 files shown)


abbyoung (approved):
LGTM! Tested locally and success 🌟

locustfiles/prime.py
```diff
@@ -27,7 +27,6 @@ class PrimeUser(MilMoveHostMixin, HttpUser):
     # These are locust HttpUser attributes that help define and shape the load test:
     wait_time = between(0.25, 9)  # the time period to wait in between tasks (in seconds, accepts decimals and 0)
     tasks = {PrimeTasks: 1}  # the set of tasks to be executed and their relative weight
-    weight = 5
```
eamahanna, line 30: Why are the weights no longer needed?

tasks/prime.py
```diff
@@ -17,49 +19,55 @@
...
 )
-
 from .base import check_response, CertTaskMixin, ParserTaskMixin
-from copy import deepcopy
 
 logger = logging.getLogger(__name__)
 
 
-def prime_path(url):
+def prime_path(url: str) -> str:
```
eamahanna, line 27: Does this syntax just mean the url has to be a string? I don't thin I have seen it before.

tasks/prime.py
```diff
@@ -426,20 +516,23 @@ def update_mto_shipment_status(self):
...
             # If this date is set here, the status will not properly transition to APPROVED
             "availableToPrimeAt": None,
             "moveOrder": {
                 "status": "APPROVED",
                 # We need these objects to exist
-                "destinationDutyStationID": "71b2cafd-7396-4265-8225-ff82be863e01",
-                "originDutyStationID": "1347d7f3-2f9a-44df-b3a5-63941dd55b34",
-                "uploadedOrdersID": "c26421b0-e4c3-446b-88f3-493bb25c1756",
+                "destinationDutyStationID": self.default_mto_ids["destinationDutyStationID"],
+                "originDutyStationID": self.default_mto_ids["originDutyStationID"],
```
eamahanna, line 534: 🎉

tasks/prime.py
sandy-wright: yep! This is called "type hinting" in Python, and it's a way we can try to make python a little stricter with data types. This line is saying that `url` needs to be a string (the `url: str` part), and it's also saying that this function will always return a string (the `-> str` part)

locustfiles/prime.py
sandy-wright: The weights were throwing off the balance of PrimeUsers vs. SupportUsers. When we still had them, Locust would spawn 5 PrimeUsers to every 1 SupportUser. This meant I'd need to spawn 6 users if I wanted the support tasks to run at all. At first I thought it made sense because we wanted to test the Prime API more than the Support API, but it just gets in the way of development now. I want to be able to run a load test with 2 users and make sure one is Prime and one is Support

tasks/prime.py
eamahanna: Should we aim to do this in all of our functions?

sandy-wright: There's no hard and fast rule one way or the other - if you have a function that only works with clearly-defined types, it would be a good idea to use type-hinting. Still, there's a lot of grey area with types in python, and there are situations where it makes less sense. I actually like to take advantage of python's loose typing though, and I know some python developers who hate that haha. 

At the very least, I would encourage using type hinting if you can. It's rarely (if ever - arguably) a bad thing, and although I don't want to make it a hard requirement, it would be nice to see more of it.

PR #53, opened 2021-02-11
MB-5938 unit test blitz

eamahanna:
## Description

This PR adds units tests for the MilMoveProvider and MilMoveData classes and their functions. 

## Reviewer Notes

My python is rusty, are there better ways to test? 

## Setup

Execute `pytest utils/tests/test_fake_data.py`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5938) for this change.

### DIFF
--- a/utils/tests/test_fake_data.py
+++ b/utils/tests/test_fake_data.py
@@ -3,9 +3,11 @@
 import pytest
 
 from utils.fake_data import MilMoveProvider
+from utils.fake_data import MilMoveData
 from faker import Faker
 from faker.generator import Generator
 from datetime import datetime
+from utils.constants import DataType
 
 
 class TestMilMoveProvider:
@@ -18,8 +20,6 @@ class TestMilMoveProvider:
         cls.fake.add_provider(MilMoveProvider)
         cls.provider = MilMoveProvider(Generator())
 
-    # faker takes functions and makes own, attributes is different
-
     def test_init(self):
         """
         Tests that the __init__ from the setup_class method worked correctly.
@@ -29,18 +29,27 @@ class TestMilMoveProvider:
         assert self.provider.last_name_used is True
 
     def test_safe_phone_number(self):
+        """
+        Tests the returned phone number is the correct length and structure
+        """
         phone_number = self.fake.safe_phone_number()
         assert type(phone_number) is str
         assert len(phone_number) == 12
         assert phone_number.index("555") == 4
 
     def test_time_military(self):
+        """
+        Tests the returned time is the correct length and structure
+        """
         time = self.fake.time_military()
         assert type(time) is str
         assert len(time) == 5
         assert time.endswith("Z")
 
     def test_iso_date_time(self):
+        """
+        Tests the returned time is a string in ISO date time format
+        """
         date_value = self.fake.iso_date_time()
         assert type(date_value) is str
         try:
@@ -50,6 +59,9 @@ class TestMilMoveProvider:
             pytest.fail(ValueError)
 
     def test__set_safe_name(self):
+        """
+        Tests the current name does not equal empty strings and sets variables to false
+        """
         self.provider._set_safe_name()
         assert self.provider.current_name["first_name"] != ""
         assert self.provider.current_name["last_name"] != ""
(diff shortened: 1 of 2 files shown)


utils/tests/test_fake_data.py
```diff
@@ -0,0 +1,122 @@
...
+        assert self.provider.last_name_used is True
+
+    def test_safe_phone_number(self):
+        """
+        Tests the returned phone number is the correct length and structure
+        """
+        phone_number = self.fake.safe_phone_number()
+        assert type(phone_number) is str
+        assert len(phone_number) == 12
+        assert phone_number.index("555") == 4
```
sandy-wright, line 38: This last assertion is a little goofy, haha. Are you comfortable with regex? This would be a good place to test that what we're getting back matches the pattern we expect to see, namely `[2-9]\d{2}-555-\d{4}`. Don't feel obligated to do pattern matching if it's new to you, but if you're curious about regex, this is a really good resource: https://regexone.com/

utils/tests/test_fake_data.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        last_name = self.fake.safe_first_name()
+        assert type(last_name) is str
+        assert last_name != self.provider.current_name["last_name"]
+
+    def test_safe_street_address(self):
+        """
+        Tests the return address is a string
+        """
+        assert type(self.fake.safe_street_address()) is str
```
sandy-wright, line 91: These tests are a little tricky because what we _really_ want to test is that the data we're getting back is the safe data from our `fake_data.json` file. This is the fake data that USTC has actually approved for our use. A check like:

```python
address = self.fake.safe_street_address()
assert address in self.provider.safe_data["addresses"]
```

For names, maybe:
```python
fake_name = self.fake.safe_first_name()
for name in self.safe_data["names"]:
    if name["first_name"] == fake_name:
        found = True
        
assert found
```

utils/tests/test_fake_data.py
```diff
@@ -0,0 +1,122 @@
...
+        assert self.provider.first_name_used is False
+        assert self.provider.last_name_used is False
+
+    def test_safe_first_name(self):
+        """
+        Tests the first name is a string and does not equal the current first name
+        """
+        first_name = self.fake.safe_first_name()
+        assert type(first_name) is str
+        assert first_name != self.provider.current_name["first_name"]
```
sandy-wright, line 77: This check right now is a little misleading - the provider you initialized as `cls.provider` in the setup is actually _not_ what's being used for generating the first name here (that's a different instance of the provider inside the Faker class). At this point, the first name actually SHOULD be the current first name, and we should see that `self.first_name_used` is True.

If we want to test this accurately, we'll have to use the MilMoveProvider class and not the Faker class. Now that we're passing in a generator, you might be able to do:
```python
self.provider.safe_first_name()
```
without an error.

utils/tests/test_fake_data.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def test_safe_last_name(self):
+        """
+        Tests the last name is a string and does not equal the current last name
+        """
+        last_name = self.fake.safe_first_name()
```
sandy-wright, line 83: Looks like a typo here - should be `self.fake.safe_last_name()` :)

utils/tests/test_fake_data.py
```diff
@@ -0,0 +1,122 @@
...
+        """ Initialize the MilMoveData that will be tested. """
+        cls.data = MilMoveData()
+
+    def test_init(self):
+        """
+        Tests that the __init__ from the setup_class method worked correctly.
+        """
+        assert isinstance(self.data.fake, Faker)
+        assert type(self.data.data_types) is dict
+        assert len(self.data.data_types) == 16
```
sandy-wright, line 108: We don't have to be that specific here, I think a good ol' `assert self.data.data_types is not None` would be sufficient.

utils/tests/test_fake_data.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def test_get_random_choice(self):
+        """
+        Tests that an element is returned from the provided list
+        """
+        list = ["string1", 4, True, "string2"]
```
sandy-wright, line 114: Hahaha, love the truly random data. One thing I would recommend is renaming `list` to `random_list` or something along those lines. `list` is an internal keyword in python, so overriding it can be very dangerous (the fact that python even lets you do this is shady as heck).

sandy-wright (changes requested):
This is really good work! I left a lot of comments, but don't feel like you need to do everything I suggested. The most important changes are the ones I mentioned in `test_safe_first_name` and `test_safe_last_name`. Beyond that, it's a lot more subjective!

Let me know if have any questions ☺️

sandy-wright (approved):
Thanks for making all those changes!! Looks great ✨

PR #54, opened 2021-02-12
[MB-6219] Add shipments and serviceitems to the prime workflow in load testing

shimonacarvalho:
## Description

This ticket adds more calls to the prime workflow. These are in the `prime_workflow.py` file.
Added workflow chunks for 
* Add shipment with doshut service and approve
* update shipment agent and address

To support that, did a bunch of cleanup and update on the tasks in `tasks/prime.py`
* Reduced logger printouts on successful calls
* Updated tasks to request a specific object from storage if called with overrides.
* Updated tasks to return the new object on successful call to the endpoint
* Documented all overrides in tasks

**This branch depends on the same named branch in the mymove repo**
Here is that PR : https://github.com/transcom/mymove/pull/5905

## Setup

- run Locust with make `load_test_prime_workflow`
- In the locust load test, select 25/10 for number of "users to simulate" and for "swarm rate".
- Check that all 7 new endpoints are hit as listed in the jira
- Check that there are ZERO failures. 

Unlike the discrete task based load testing, the prime_workflow taskset will **only test happy paths**. Therefore we expect no failures.
You should see hits to the above endpoints. You should see ZERO failures.

There are also tests setup for the storage aspect of prime workflow
You can run them with 
`pytest tasks/tests/test_prime_workflow.py` 

Should also have ZERO failures.

## References

* [Jira story](https://dp3.atlassian.net/browse/mb-6219) for this change.

### DIFF
--- a/tasks/base.py
+++ b/tasks/base.py
@@ -21,7 +21,7 @@ def check_response(response: Response, task_name="Task", request=None):
     :param request: any type, optional data to print for debugging a failed response
     :return: tuple(dict, bool)
     """
-    logger.info(f"ℹ️ {task_name} status code: {response.status_code}")
+    logger.info(f"ℹ️ {task_name} status code: {response.status_code} {response.reason}")
 
     try:
         json_response = json.loads(response.content)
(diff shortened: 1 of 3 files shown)


eamahanna (approved):
LGTM and works correctly. Verified all the actions were completed in workflow without error

PR #58, opened 2021-02-25
Sc mb 5939 tests for utils hosts

eamahanna:
## Description

This PR adds test for `utils/hosts.py`. 

## Reviewer Notes

Is there a better way to handle the `@mock.patch.dict(os.environ, {"MOVE_MIL_EXP_TLS_CERT": "test_cert", "MOVE_MIL_EXP_TLS_KEY": "test_key"})` calls? 

Did I test the `TestMilMoveDomain` function correctly?

## Setup

To run the tests, execute `pytest utils/tests/test_hosts.py`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5939) for this change.


utils/tests/test_hosts.py
```diff
@@ -0,0 +1,162 @@
...
+        """ Define and initialize classes to be tested """
+
+        class HostLocal(MilMoveHostMixin):
+            # These attributes are used in MilMoveHostMixin to set up the proper hostname for any MilMove environment:
+            local_port = "9443"
+            domain = MilMoveDomain.PRIME  # the base domain for the host
+            is_api = True  # if True, uses the api base domain in deployed environments
+            host = "local"
+
+        """ Initialize the APIParser that will be tested. """
```
sandy-wright, line 45: In general, it's good to avoid using docstring comments unless they are documenting a class or a function definition. The reason for this is because they aren't ignored by the compiler like normal comments, so it creates unnecessary noise when you run the code.

utils/tests/test_hosts.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class TestMilMoveHostMixin:
+    """ Tests the MilMoveHostMixin class and its methods. """
+
+    @classmethod
+    def setup_class(cls):
+        """ Define and initialize classes to be tested """
+
+        class HostLocal(MilMoveHostMixin):
```
sandy-wright, line 38: I might suggest renaming this to something like `TestHostUser` instead of `HostLocal`. It's not _always_ going to be local during the tests, plus this obfuscates the fact that this is meant to be used like an `HttpUser` class. BUT this is definitely up to you, I'm fine with it if you keep it as-is too

utils/tests/test_hosts.py
```diff
@@ -0,0 +1,162 @@
...
+        assert re.search("static/tls/devlocal-mtls.cer", self.TestUser1.cert_kwargs["cert"][0])
+        assert re.search("static/tls/devlocal-mtls.key", self.TestUser1.cert_kwargs["cert"][1])
+
+    def test_set_milmove_env(self):
+        self.HostLocalClass.env = None
+        self.TestUser1.set_milmove_env("exp")
+
+        assert self.HostLocalClass.env == MilMoveEnv.EXP
+        assert self.TestUser1.env == MilMoveEnv.EXP
+        assert self.TestUser2.env == MilMoveEnv.EXP
```
sandy-wright, line 65: nice!

utils/tests/test_hosts.py
```diff
@@ -0,0 +1,162 @@
...
+        Tests the host name for the class can be set using the class env
+        and be called from a class instance
+        """
+        # reset all initally set values in class
+        self.HostLocalClass.host = None
+        self.HostLocalClass.env = MilMoveEnv.EXP
+
+        # host is None due to setting the class host to None above
+        assert self.TestUser1.host is None
+        assert self.TestUser2.host is None
```
sandy-wright, line 84: 💯 ✨

utils/tests/test_hosts.py
```diff
@@ -0,0 +1,162 @@
...
+        self.HostLocalClass.env = MilMoveEnv.LOCAL
+        self.TestUser1.set_cert_kwargs()
+
+        assert re.search("tls/devlocal-mtls.cer", self.TestUser1.cert_kwargs["cert"][0])
+        assert re.search("tls/devlocal-mtls.key", self.TestUser1.cert_kwargs["cert"][1])
+        assert re.search("tls/devlocal-mtls.cer", self.TestUser2.cert_kwargs["cert"][0])
+        assert re.search("tls/devlocal-mtls.key", self.TestUser2.cert_kwargs["cert"][1])
+
+    @mock.patch.dict(os.environ, {"MOVE_MIL_EXP_TLS_CERT": "test_cert", "MOVE_MIL_EXP_TLS_KEY": "test_key"})
+    def test_create_deokoyed_cert_file(self):
```
sandy-wright, line 112: I think this should be "deployed" instead of "deokoyed" 😄

utils/tests/test_hosts.py
```diff
@@ -0,0 +1,162 @@
...
+        self.HostLocalClass.env = MilMoveEnv.EXP
+        cert_file_path = self.TestUser1.create_deployed_cert_file()
+        test_file_contents = "test_cert\ntest_key"
+
+        assert re.search("tls/exp_tls_cert_key.pem", cert_file_path)
+        with open(cert_file_path, "r") as f:
+            assert f.read() == test_file_contents
+
+    def test_no_env_variables_create_deokoyed_cert_file(self):
```
sandy-wright, line 121: same typo as above, I love how it's clear that your right hand was just one key too far left haha

utils/tests/test_hosts.py
```diff
@@ -0,0 +1,162 @@
...
+
+    @mock.patch.dict(os.environ, {"MOVE_MIL_EXP_TLS_CERT": "test_cert", "MOVE_MIL_EXP_TLS_KEY": "test_key"})
+    def test_create_deokoyed_cert_file(self):
+        self.HostLocalClass.env = MilMoveEnv.EXP
+        cert_file_path = self.TestUser1.create_deployed_cert_file()
+        test_file_contents = "test_cert\ntest_key"
+
+        assert re.search("tls/exp_tls_cert_key.pem", cert_file_path)
+        with open(cert_file_path, "r") as f:
+            assert f.read() == test_file_contents
```
sandy-wright, line 119: nice, this is a great test

utils/tests/test_hosts.py
```diff
@@ -0,0 +1,162 @@
...
+    def test_remove_deployed_cert_file(self):
+        # Setup kwargs
+        self.HostLocalClass.env = MilMoveEnv.EXP
+        self.HostLocalClass.cert_kwargs = None
+        self.TestUser1.set_cert_kwargs()
+        cert_kwargs_before = self.TestUser1.cert_kwargs["cert"]
+
+        self.TestUser1.remove_deployed_cert_file()
+        assert os.path.exists(cert_kwargs_before) is False
+        assert self.TestUser1.cert_kwargs == {}
```
sandy-wright, line 136: Can we double-check that the `cert_kwargs == {}` for TestUser2 as well?

utils/tests/test_hosts.py
```diff
@@ -0,0 +1,162 @@
...
+
+        # Call function again to make sure it doesn't error
+        self.TestUser1.remove_deployed_cert_file()
+
+
+# tasks.base logger for mocking
+logger = logging.getLogger("utils.hosts")
```
sandy-wright, line 143: Not a big deal, but you could put this on the top of the file. OR you could put it within the `test_clean_milmove_host_users` function to make it clear that's the only place where it's used. Either way though, doesn't really matter

utils/tests/test_hosts.py
```diff
@@ -0,0 +1,162 @@
...
+        test_file_contents = "test_cert\ntest_key"
+
+        assert re.search("tls/exp_tls_cert_key.pem", cert_file_path)
+        with open(cert_file_path, "r") as f:
+            assert f.read() == test_file_contents
+
+    def test_no_env_variables_create_deokoyed_cert_file(self):
+        self.HostLocalClass.env = MilMoveEnv.EXP
+        with pytest.raises(ImplementationError):
+            self.TestUser1.create_deployed_cert_file()
```
sandy-wright, line 124: Hmm, this test doesn't work for me because it's finding my actually environment variables. Maybe you should add the mock to this method, but with an empty dict? Like `@mock.patch.dict(os.environ, {})`? Or we could omit this test if necessary. Didn't realize my actual env vars would goof this up, but I guess it makes sense 🤔

sandy-wright (changes requested):
These tests look great!! Really good test cases in here. Only thing that should be changed is the `test_no_env_variables_create_deokoyed_cert_file` test on line 121 - it fails for me because it finds the actual EXP environment variables in my shell env. That can be mocked out, or remove the test if necessary, whichever works best for you.

utils/tests/test_hosts.py
eamahanna: LOL

utils/tests/test_hosts.py
eamahanna: At least it was consistent lol

sandy-wright (approved):
LGTM! 🚀

utils/tests/test_hosts.py
eamahanna: Fixed!

PR #59, opened 2021-02-26
MB-5958 Add unit tests for utils.fields.ObjectField

sandy-wright:
## Description

This PR adds unit tests for all of the methods in the ObjectField class. It also fixes some bugs encountered along the way.

## Setup

To run only the new tests:

```sh
pytest -v -k TestObjectField utils/tests/test_fields.py
```

And to run all tests:

```sh
pytest
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5958) for this change.

### DIFF
--- a/utils/tests/test_fields.py
+++ b/utils/tests/test_fields.py
@@ -3,11 +3,15 @@
 import re
 
 import pytest
+
+# from pytest_mock import mocker
 from faker.generator import Generator
 
 from utils.fields import BaseAPIField, ArrayField, EnumField, ObjectField
 from utils.constants import DataType
 from utils.fake_data import MilMoveProvider, MilMoveData
+
+MOCK_FAKE_DATA = {DataType.INTEGER: 112358}
 
 
 class TestBaseAPIField:
@@ -80,6 +84,7 @@ class TestObjectField:
     def setup_class(cls):
         """ Initialize the ObjectField that will be tested. """
         cls.object_field = ObjectField()
+        cls.faker = MilMoveData()
 
     def test_init(self):
         """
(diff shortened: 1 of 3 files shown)


utils/tests/test_fields.py
```diff
@@ -69,3 +73,324 @@ def test_generate_fake_data(self):
...
+        assert test_field in self.object_field.object_fields
+
+        # Testing that we can add another ObjectField to this ObjectField's list of subfields,
+        # and that the subfields of the two Objects do not merge:
+        test_object = ObjectField(name="testObject")
+        test_object_subfield = BaseAPIField(name="testSentence", data_type=DataType.SENTENCE)
+        test_object.add_field(test_object_subfield)
+        self.object_field.add_field(test_object)
+        assert test_object in self.object_field.object_fields
+        assert test_object_subfield in test_object.object_fields
```
eamahanna, line 114: Nice 🎉

utils/tests/test_fields.py
```diff
@@ -69,3 +73,324 @@ def test_generate_fake_data(self):
...
+        self.object_field.combine_fields(array_field, unique=True)
+        assert len([field for field in self.object_field.object_fields if field.name == "arrayField"]) == 1
+
+        test_object = ObjectField(name="testObject")
+        test_object_subfield = BaseAPIField(name="testSentence", data_type=DataType.SENTENCE)
+        test_object.add_field(test_object_subfield)
+
+        self.object_field.combine_fields(test_object)
+        assert test_object not in self.object_field.object_fields
+        assert all([test_field in self.object_field.object_fields for test_field in test_object.object_fields])
```
eamahanna, line 175: Can you explain this check?

utils/tests/test_fields.py
```diff
@@ -69,3 +73,324 @@ def test_generate_fake_data(self):
...
+        assert base_field.required is False
+        assert random_field.required is False
+        assert enum_field.required is False
+        assert space_field.required is False
+
+        self.object_field.update_required_fields(required_fields)
+
+        assert base_field.required is True
+        assert random_field.required is True
+        assert enum_field.required is False
```
eamahanna, line 211: Why is this still false?

utils/tests/test_fields.py
sandy-wright: Yes (and I'll add a comment too):

The main purpose for the `ObjectField.combine_fields` method is to combine the fields from two distinct ObjectFields into one ObjectField. To simplify:

```
ObjectA fields:
  - name
  - date_of_birth
  
ObjectB fields:
  - favorite_color
  
ObjectA.combine_fields(ObjectB)

ObjectA fields:
  - name
  - date_of_birth
  - favorite_color
```

So this test is trying to make sure that every field that was originally in `test_object`'s list of fields is now _also_ in `self.object_field`'s list of fields. To do that, we loop through each field in `test_object.object_fields`, check if that field is also in `self.object_field.object_fields`, and then add the Boolean result of that check to a list. At the end we're checking that the list is all True - meaning every single field in `test_object` was added to `self.object_field`.

It definitely doesn't help that everything has "object" and "field" in the name 😩 . I might rename `test_object` to help with that

utils/tests/test_fields.py
sandy-wright: Because its name wasn't in the `required_fields` list I passed in. Only these fields were going to be marked required: `["baseField", "randomField", "phoneField", "spaceField"]`. I set that at the top of this function, which I don't think is very readable. I'll change it to look like:

```python
        self.object_field.update_required_fields(["baseField", "randomField", "phoneField", "spaceField"])

        assert base_field.required is True
        assert random_field.required is True
        assert enum_field.required is False
        assert space_field.required is True
```

Hopefully that makes it more readable

eamahanna (approved):
LGTM :)

PR #61, opened 2021-03-17
MB-6698 Fix updateMTOServiceItem 500 errors during load testing

eamahanna:
## Description

Previously, load testing would generate random UUID's for tests. This led to foreign key errors being through when calling `updateMTOServiceItem` and `createMTOServiceItem`. To fix this, the faker now returns an empty uuid. The empty uuid is ignored by swagger/handlers and they won't be updated now. 

## Setup

To test that this works, run load testing and confirm that no 500 errors appear for 
`PATCH | /prime/v1/mto-service-items/{mtoServiceItemID}` and `POST | /prime/v1/mto-service-items`.

To run load testing run `make db_dev_e2e_populate` and `make server_run` in the mymove repo. Then run `make load_test_prime` in the load testing repo.

## References

* [MB-6698](https://dp3.atlassian.net/browse/MB-6698) for this change.
* [MB-6697](https://dp3.atlassian.net/browse/MB-6697) for this change.

### DIFF
--- a/utils/fake_data.py
+++ b/utils/fake_data.py
@@ -95,6 +95,12 @@ class MilMoveProvider(AddressProvider, DateProvider):
             if state != "AK" and state != "HI":
                 return self.postalcode_in_state(state)
 
+    def safe_uuid(self):
+        """
+        Returns an empty uuid as a string.
+        """
+        return "00000000-0000-0000-0000-000000000000"
+
 
 class MilMoveData:
     """ Base class to return fake data to use in MilMove endpoints. """
@@ -121,7 +127,7 @@ class MilMoveData:
             DataType.SENTENCE: self.fake.sentence,
             DataType.BOOLEAN: self.fake.boolean,
             DataType.INTEGER: self.fake.random_number,
-            DataType.UUID: self.fake.uuid4,
+            DataType.UUID: self.fake.safe_uuid,
         }
 
     def get_random_choice(self, choices):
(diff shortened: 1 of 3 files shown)


utils/fake_data.py
```diff
@@ -95,6 +95,12 @@ def safe_postal_code(self):
             if state != "AK" and state != "HI":
                 return self.postalcode_in_state(state)
 
+    def safe_uuid(self):
+        """
+        Returns an empty uuid as a string.
+        """
+        return "00000000-0000-0000-0000-000000000000"
```
eamahanna, line 102: Is there a better way to do this?

sandy-wright: Yep! There's a constant for this value, so add `ZERO_UUID` to the imports from constants on line 11, and then you can do `return ZERO_UUID` here

sandy-wright (commented):
This looks good, and mostly fixes the issue! I noticed there were still some 500 errors with the `update_mto_service_item` task, so I poked around a little and found the problem.

If you look at line 481 of `tasks/prime.py`, you'll see that the overrides passed in are just `overrides={"id": mto_service_item["id"]}` - this sets _all_ "id" fields to the service item's ID, including the address! So to solve this, I tried this jumble of code:

```python
payload = self.fake_request(
    "/mto-service-items/{mtoServiceItemID}",
    "patch",
     overrides={
           "id": mto_service_item["id"],
           "sitDestinationFinalAddress": {
               "id": mto_service_item["sitDestinationFinalAddress"]["id"] if
                mto_service_item.get("sitDestinationFinalAddress") and
                mto_service_item["sitDestinationFinalAddress"].get("id") else ZERO_UUID,
        }
    }
)
```

This says "If there's a `sitDestinationFinalAddress` on this shipment and it has an `id` value, use that `id`. Otherwise, use the zero uuid." Let me know if you have any questions about that!

sandy-wright (approved):
LGTM!

PR #62, opened 2021-03-18
MB-7225 Stop generating negative crate dimensions

sandy-wright:
## Description

For the Domestic Crating MTO Service Items, the API parser has to generate a set of dimensions for the item and a set of dimensions for the crate that item is going to be put in. These values are all random integers, but we have to make sure the item is smaller than the crate for the endpoint to work.

With the old code, we would just set the item value to be the same as the crate's - 1 so that we knew it was smaller:
```python
if item <= crate:
    item = crate - 1
```

However, in rare cases, the generator would pick 0 as a value for one of the crate's dimensions! In that case the item, would get -1! We don't want to break the rules of space and time with this code, so I've changed the above code to also make sure both values are above 0.

## Setup

To test this in action, run load testing for a while and verify that you don't see ANY 500 errors from `createMTOServiceItem`. You may also run the unit tests for this function:

```sh
pytest -v -k TestPrimeAPIParser
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7225) for this change.

### DIFF
--- a/utils/parsers.py
+++ b/utils/parsers.py
@@ -435,7 +435,7 @@ class PrimeAPIParser(APIParser):
         :return: tup(item_dim int, crate_dim int)
         """
         if item_dim <= 0:
-            item_dim += 1
+            item_dim = 1
         if item_dim >= crate_dim:
             if crate_dim <= 1:
                 crate_dim = 2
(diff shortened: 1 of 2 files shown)


shimonacarvalho (approved):
LGTM! Tested.

PR #66, opened 2021-05-13
[MB-7254] GHC API load testing getMove endpoint for SC and TOO

duncan-truss:
## Description

This PR expands the GHC API testing to incorporate distinct Services Counselor and TOO office users making requests to the [getMove endpoint](http://milmovelocal:3000/swagger-ui/ghc.html#/move/getMove).

I incorporated the same kind of local storage that the Prime uses as a mixin to both set of office user tasks.  I added tasks to request the queue data for each user type, which will populate the local storage with move data in different statuses.  There is overlap from the Services Counselor moves and TOO moves for those in the `SERVICE COUNSELING COMPLETED` status so what ends up in the local store will naively contain duplicates at this time.

To make things more realistic, I did work on randomizing creation of move data for the services counselor user because the seed only had 1 or 2 moves.  That [PR will also need review in the mymove repo](https://github.com/transcom/mymove/pull/6624).  For my load test I bumped up the seed data to create ~200 moves.

While simulating hundreds of users I did start to see failures with a server error of too many open files.  There is a default system/session soft limit of 256 open files.  You can see this by running `ulimit -a` in your terminal.  Once I raised it with with `ulimit -n 2048` I now longer saw failures when load testing a high number of users.  This limit may be per terminal tab so it may not apply across sessions, so set it in the window where you want to run your server.

## Reviewer Notes

We should likely be seeing some failures if we had authorization on this endpoint to make sure a role such as services counselor can not view a move in say Approved or Approvals Requested or a TOO cannot view a move in Needs Service Counseling.

Both SC and TOO users are in the same default GBLOC so there is no chance of requesting one outside but we should try to introduce that maybe by adding USMC users too.

We could also introduce non-existent locator requests, the current exception now is just when no moves exist yet in the local store lookup before a queue response is added.

## Setup

From the [new branch in mymove](https://github.com/transcom/mymove/pull/6624) with the devseed changes, editing the loop count to create the desired amount of moves.
```sh
make db_dev_e2e_populate
make server_run
```

From the milmove_load_testing repo:
```sh
make local_docker_build
make local_docer_office_up
```

In locust I simulated 500 users that spawned 5 new users/per second.

Screen Shot 2021-05-14 at 11 14 47 AM
Screen Shot 2021-05-14 at 11 14 55 AM
Screen Shot 2021-05-14 at 11 15 03 AM

[Downloadable html report (must be zipped for github upload) ](https://github.com/transcom/milmove_load_testing/files/6479649/report_1621002704.280355.html.zip)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7254) for this change.

### DIFF
--- a/tasks/office.py
+++ b/tasks/office.py
@@ -32,7 +32,7 @@ class OfficeDataStorageMixin:
     This mixin allows storing multiple items of each kind.
     """
 
-    DATA_LIST_MAX: int = 50
+    DATA_LIST_MAX: int = 100
     local_store: Dict[str, list] = {
         MOVE_TASK_ORDER: [],
         MTO_SHIPMENT: [],
@@ -45,7 +45,7 @@ class OfficeDataStorageMixin:
         Given an object_key that represents an object type from the MilMove app, returns an object of that type from the
         list.
 
-        :param object_key: str in [MOVE_TASK_ORDER, MTO_SHIPMENT, MTO_AGENT, MTO_SERVICE_ITEM, PAYMENT_REQUEST]
+        :param object_key: str in [MOVE_TASK_ORDER, MTO_SHIPMENT, MTO_SERVICE_ITEM, PAYMENT_REQUEST]
         """
         data_list = self.local_store[object_key]
 
@@ -63,10 +63,9 @@ class OfficeDataStorageMixin:
         :return: None
         """
         data_list = self.local_store[object_key]
-        logger.info(f"size of {object_key} {len(data_list)}")
 
         if len(data_list) >= self.DATA_LIST_MAX:
-            num_to_delete = random.randint(1, self.DATA_LIST_MAX)
+            num_to_delete = random.randint(1, len(data_list))
             del data_list[:num_to_delete]
 
         # Some creation endpoint auto-create multiple objects and return an array,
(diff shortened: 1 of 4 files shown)


donaldthai (approved):
I have limited Python experience but I was able to get everything to run and code looks okay to me :) Maybe @sandy-wright should take a look before merging though!

tasks/office.py
```diff
@@ -15,9 +24,81 @@ def ghc_path(url: str) -> str:
...
+        elements, and if so, it randomly removes 1 to MAX number of elements so that the cycle can start again (and so
+        we don't hog too much memory).
+
+        :param object_key: str in [MOVE_TASK_ORDER, MTO_SHIPMENT, MTO_AGENT, MTO_SERVICE_ITEM, PAYMENT_REQUEST]
+        :param object_data: JSON/dict
+        :return: None
+        """
+        data_list = self.local_store[object_key]
+
+        if len(data_list) >= self.DATA_LIST_MAX:
```
Ryan-Koch, line 67: Nice way to keep the list from ballooning

Ryan-Koch (approved):
It's been a minute since I got to review some python, but it looks solid to me.

duncan-truss:
> It's been a minute since I last got to review some python, but it looks solid to me.

I didn't write much custom Python here, the local storage mixin at this stage is almost exactly how the prime storage behaves.

PR #70, opened 2021-05-19
MB-7939 test create_mto_shipments endpoint

eamahanna:
## Description

This PR adds a new file `prime_endpoint_workflows.py`. The intent of this file is to create workflows that will specifically target a single endpoint. To test the primeapi create_mto_shipment endpoint, run the following command:
 
`locust -f locustfiles/prime_workflow --tags createMTOShipmentWorkflow`

I also created added the first two pages to this repo's wiki 🎉 and updated the readme to point to the wiki for some custom test command information.

https://github.com/transcom/milmove_load_testing/wiki

## Reviewer Notes

The naming of the files is a little wonky. Do you have any better suggestions?

## Setup

To load test the create_mto_shipment endpoint complete the following steps:
1. `make server_run` in the mymove repo
2. `locust -f locustfiles/prime_workflow --tags createMTOShipmentWorkflow` in the load testing repo
3. Enter a set number of users and spawn rate (It's easy to test with 10 and 10)

## References

* [Jira story](tbd) for this change.
* [This article](tbd) explains more about the approach used.

### DIFF
--- a/locustfiles/prime_workflow.py
+++ b/locustfiles/prime_workflow.py
@@ -4,6 +4,7 @@ from locust import HttpUser, between
 from utils.hosts import MilMoveHostMixin, MilMoveDomain
 from utils.constants import PRIME_API_KEY, SUPPORT_API_KEY
 from tasks import PrimeWorkflowTasks
+from tasks.prime_endpoint_workflows import PrimeEndpointWorkflowsTasks
 from utils.parsers import SupportAPIParser, PrimeAPIParser
 
 support_api = SupportAPIParser()
@@ -24,5 +25,5 @@ class PrimeWorkflowUser(MilMoveHostMixin, HttpUser):
     parser = {PRIME_API_KEY: prime_api, SUPPORT_API_KEY: support_api}
 
     # For the tasks, we can define all the workflow tasksets we have set up:
-    tasks = {PrimeWorkflowTasks: 1}
+    tasks = {PrimeWorkflowTasks: 1, PrimeEndpointWorkflowsTasks: 1}
     weight = 1
--- a/locustfiles/prime_workflow.py
+++ b/locustfiles/prime_workflow.py
@@ -3,7 +3,7 @@ from locust import HttpUser, between
 
 from utils.hosts import MilMoveHostMixin, MilMoveDomain
 from utils.constants import PRIME_API_KEY, SUPPORT_API_KEY
-from tasks import PrimeWorkflowTasks
+from tasks.prime_hhg_workflow import PrimeWorkflowTasks
 from tasks.prime_endpoint_workflows import PrimeEndpointWorkflowsTasks
 from utils.parsers import SupportAPIParser, PrimeAPIParser
 
--- a/tasks/prime_endpoint_workflows.py
+++ b/tasks/prime_endpoint_workflows.py
@@ -9,7 +9,7 @@ from utils.constants import MOVE_TASK_ORDER
 logger = logging.getLogger(__name__)
 
 
-@tag("workflow")
+@tag("endpointWorkflow")
 class PrimeEndpointWorkflowsTasks(PrimeTasks, SupportTasks):
     """Workflow Task Set
     Each task is a workflow from start to finish.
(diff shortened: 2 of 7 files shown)


tasks/prime_endpoint_workflows.py
```diff
@@ -0,0 +1,50 @@
...
+        This is a basic workflow that checks if there is an existing move, then calls the create_mto_shipment endpoint.
+        """
+        self.workflow_title = "create_mto_shipment endpoint"
+
+        # Ensure there is a move avaliable for a shipment
+        if not self._get_move():
+            return
+
+        # Set shipment type to HHG, so there can be multiple shipments on the same move
+        overrides = {"shipmentType": "HHG"}
```
eamahanna, line 37: @sandy-wright should I leave this override or let it fail once an NTS shipment has been added once? It is suppose to fail in that scenario....

tasks/prime_endpoint_workflows.py
```diff
@@ -0,0 +1,50 @@
...
+
+from locust import tag, task
+
+from .prime import PrimeTasks, SupportTasks
+from utils.constants import MOVE_TASK_ORDER
+
+logger = logging.getLogger(__name__)
+
+
+@tag("workflow")
```
sandy-wright, line 12: I might make this workflow tag different from the other one, just so it can also be used to differentiate from the other workflow. But, frankly, I don't know if it's necessary - we'll probably be using the tags on the tasks themselves to do that

tasks/prime_endpoint_workflows.py
sandy-wright: I would go ahead and leave it, these workflows are supposed to work 100% of the time. The comment is very helpful too!

tasks/prime_endpoint_workflows.py
```diff
@@ -0,0 +1,50 @@
...
+        overrides = {"shipmentType": "HHG"}
+        shipment = super().create_mto_shipment(overrides)
+        logger.info(f"{self.workflow_title} - Created shipment {shipment['id'][:8]}")
+
+    def _get_move(self):
+        move_task_order = super().get_stored(MOVE_TASK_ORDER)
+        if not move_task_order:
+            if not self.has_all_default_mto_ids():
+                super().fetch_mto_updates()
+            move_task_order = super().create_move_task_order()
```
sandy-wright, line 46: You should use `self` instead of `super()` for these function calls. `super()` is only needed if you're overriding a function on a class and want to call the version of the function from the parent class(es). For example:
```python
class Student(Person):

def __init__(name, birthday, classes): # overrides the __init__ method from Person
    super().__init__(name, birthday) # calls the __init__ method from Person
    self.classes = classes
```

You're not actually overriding anything here (the function names have to match exactly for that), so you just want to call the methods that are on this class already (because they were inherited). So `self` is good here.

sandy-wright (commented):
LGTM, although I haven't gotten the chance to test it yet. Left some comments, and will test it ASAP once load testing is done!

tasks/prime_endpoint_workflows.py
eamahanna: Ahhhh, noted

locustfiles/prime_workflow.py
```diff
@@ -3,7 +3,8 @@
 
 from utils.hosts import MilMoveHostMixin, MilMoveDomain
 from utils.constants import PRIME_API_KEY, SUPPORT_API_KEY
-from tasks import PrimeWorkflowTasks
+from tasks.prime_hhg_workflow import PrimeWorkflowTasks
+from tasks.prime_endpoint_workflows import PrimeEndpointWorkflowsTasks
```
sandy-wright, line 7: Oh, this is super minor, but since you put the imports for these into the `tasks/__init__.py`, you can actually simplify this import to `from tasks import PrimeWorkflowTasks, PrimeEndpointWorkflowsTasks`

sandy-wright (approved):
Works well! 🎉 

Left another small comment about code formatting, but looks good overall. The names are awkward, but I have no suggestion for improvements lol. Maybe use the singular for workflow: `PrimeEndpointWorkflowTasks`? Idk

PR #71, opened 2021-05-20
[MB-8067] Load testing GHC APIs for SC & TOO move details orders, shipments, and service items 

duncan-truss:
## Description

This PR adds load testing tasks to the office users requesting the GHC API for orders, shipments, and service items.  The TOO user will make request for all of this data on the Move Details page whereas the Services Counselor will do the same except for service items, which will not exist.

The other big change in this PR is that I've change the local store to save moves and other data in a normalized Dict instead of a list.  With the former list we would likely be filling the local store with duplicate data and had to search if we want to update a particular item.  This will come in handy as we can make paginated requests for the queue data and it will only add the non-existent data and not overwrite.  Then when an individual record is fetched with the full details, we can update that single record that previously only had cursory data from the queue results.

Load testing report for simulating 500 office users w/ 1000 moves in the local database: 
[report_1621540220.4369168.html.zip](https://github.com/transcom/milmove_load_testing/files/6518531/report_1621540220.4369168.html.zip)

PR in the mymove repository with payload changes to the queue and authorization: https://github.com/transcom/mymove/pull/6669

## Setup

```sh
make local_docker_build
make local_docker_office_up
```

If you are testing a significant number of users > 200 then you should increase the file descriptor limit where you run `make server_run`.  To raise the limit from 256 you would use the command `ulimit -n 2048`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8067) for this change.

## Screenshots

Screen Shot 2021-05-20 at 4 07 13 PM
Screen Shot 2021-05-20 at 4 07 23 PM
Screen Shot 2021-05-20 at 4 07 33 PM
Screen Shot 2021-05-20 at 4 07 42 PM


tasks/office.py
```diff
@@ -33,30 +34,36 @@ class OfficeDataStorageMixin:
...
         list.
 
-        :param object_key: str in [MOVE_TASK_ORDER, MTO_SHIPMENT, MTO_SERVICE_ITEM, PAYMENT_REQUEST]
+        :param object_key: str in [MOVE_TASK_ORDER, MTO_SHIPMENT, MTO_SERVICE_ITEM, ORDER, PAYMENT_REQUEST]
+        :param object_id: str uuid of a single object item key to return
         """
         data_list = self.local_store[object_key]
+        data_item = data_list.get("object_id")
+        if data_item is not None:
+            return data_item
```
sandy-wright, line 56: Should this be `data_item = data_list.get(object_id)`? This is trying to get an item with the literal key `"object_id"` instead of whatever value the variable `object_id` has. I was under the impression the shape of this dict would be like:
```python
{
    "uuid1": {<object data here>},
    "uuid2": {<object 2 data here>},
}
```
but I may have misunderstood.

tasks/office.py
```diff
@@ -66,34 +73,16 @@ def add_stored(self, object_key, object_data):
 
         if len(data_list) >= self.DATA_LIST_MAX:
             num_to_delete = random.randint(1, len(data_list))
-            del data_list[:num_to_delete]
+            # Convert a dict to a list so we can take a slice by insertion order fifo
+            self.local_store[object_key] = dict(list(data_list.items())[num_to_delete:])
```
sandy-wright, line 77: This is an interesting edit! We might want to do `num_to_delete - 1` because this would cause an exception if it happened to be the value of the length of the list (that's my bad originally).

tasks/office.py
```diff
@@ -66,34 +73,16 @@ def add_stored(self, object_key, object_data):
...
 
         # Some creation endpoint auto-create multiple objects and return an array,
         # but each object in the array should still be considered individually here:
         if isinstance(object_data, list):
-            data_list.extend(object_data)
+            normalized = {value["id"]: value for value in object_data}
```
sandy-wright, line 82: Oooh I don't see this syntax often. It might be worth adding a comment about what's happening here.

tasks/office.py
```diff
@@ -66,34 +73,16 @@ def add_stored(self, object_key, object_data):
...
+            self.local_store[object_key] = {**normalized, **data_list}
```
sandy-wright, line 83: The use of `data_list` is starting to get confusing because it's not actually a list anymore. Would you be willing to change it to `data_dict` or something else that's a bit more informative?

sandy-wright: This is also unusual syntax and might be worth a comment. It is neat though. I think this mixin needs unit tests as well 🤔  (although you don't need to take point on adding those now - I also think we should standardize and combine this with the Prime storage mixin)

tasks/office.py
```diff
@@ -142,6 +135,9 @@ def get_move(self, overrides=None):
             **self.user.cert_kwargs,
         )
         new_mto, success = check_response(resp, "getMove")
+        if success:
+            self.add_stored(MOVE_TASK_ORDER, new_mto)
+        return new_mto
```
sandy-wright, line 140: Should this be indented, or do you want to return the payload regardless of if it was a success or not? I think in the Prime we just return None

tasks/office.py
```diff
@@ -163,9 +159,61 @@ def get_moves_queue(self, overrides=None):
...
+        # If id was provided, get that specific one. Else get any stored one.
+        object_id = overrides.get("id") if overrides else None
+        move = self.get_stored(MOVE_TASK_ORDER, object_id)
+        if move is None:
+            logger.info("Skipping get orders no move is stored yet")
+            return
+
+        orders_id = move.get("ordersId")
+        if orders_id is None:
+            move = self.get_move({"id": move["id"]})
```
sandy-wright, line 181: It feels weird to see a dict passed in like this instead of a keyword argument, but I guess it's not wrong? Not really sure where I land on that

tasks/office.py
```diff
@@ -215,6 +266,9 @@ def get_move(self, overrides=None):
             **self.user.cert_kwargs,
         )
         new_mto, success = check_response(resp, "getMove")
+        if success:
+            self.add_stored(MOVE_TASK_ORDER, new_mto)
+        return new_mto
```
sandy-wright, line 271: Same question as above

tasks/office.py
```diff
@@ -232,5 +286,81 @@ def get_moves_queue(self, overrides=None):
             **self.user.cert_kwargs,
         )
         moves, success = check_response(resp, "getMovesQueue")
+        if success:
+            self.add_stored(MOVE_TASK_ORDER, moves["queueMoves"])
 
-        self.add_stored(MOVE_TASK_ORDER, moves["queueMoves"])
+    @tag(ORDER, "getOrder")
+    @task
+    def get_orders(self, overrides=None):
```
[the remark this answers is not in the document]
sandy-wright: +1 that we should think about that. It's tricky though, might need a good brainstorming sesh. We need to remember that we don't have a lot of folks on MM who are super familiar with Python, so we do want to avoid any solutions that are too complex or abstract

sandy-wright (changes requested):
Looks like there might be a couple of issues in the storage mixin, and I had a couple of other questions throughout. Great work though! It's really cool to see folks picking up this codebase so well 😄

tasks/office.py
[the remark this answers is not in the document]
Ryan-Koch: +1 on the idea of doing some brainstorming on that. I suspect there's a happy place in the abstraction spectrum we can shoot for.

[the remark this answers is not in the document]
duncan-truss: Agreed, I'm thinking we may still need to be smarter about how we store things by status and that is a looming thing that may influence how we do things.

tasks/office.py
duncan-truss: It's technically a dict comprehension but happy to add a comment too about the transformation.

tasks/office.py
duncan-truss: Agreed, I added the comment to the top of the function but can talk about the combination here for which keys get overwritten.  Tests are a great idea and would give me more confidence as my Python brain is a little rusty.

tasks/office.py
duncan-truss: I don't think I meant for it to be indented, I don't feel that strongly about it though.  I don't think Python has a strict a convention on return values as Go because of the error return type.

tasks/office.py
duncan-truss: That was your overrides functionality if I'm understanding it correctly.  Are you saying we don't need it to be so generalized?

tasks/office.py
sandy-wright: Maybe one comment for both of these lines that these are different examples of dict comprehension/composition. It's a neat example of two distinct strategies to do this!

tasks/office.py
sandy-wright: Tests could be a separate ticket imo! That's what Roci did, since folks had to get used to the testing environment. I'd be happy to pair on them too

tasks/office.py
sandy-wright: OH, I was confused, I thought this was some kind of `get_stored` call. But it's an endpoint! Blah, sorry 😄

tasks/office.py
duncan-truss: Done!

tasks/office.py
sandy-wright: The reason we might want to indent this is because receiving a `None` from this function would be a clear way to indicate that the endpoint call was not successful. I _think_ in the Prime tasks, we assume that if we get a JSON response, that it is the JSON for a 200 response. 

It shouldn't cause an issue if you're handling the returned values carefully (and I think you are), but I guess I just like returning None as a way to indicate failure. I really don't have that strong of an opinion about this either, so whichever way feels best to you

sandy-wright (approved):
LGTM, thanks for fixing the important issues! If you do anything else, I would like renaming `data_list` to `data_dict`, but nothing else is required ☺️ 

I'm also curious, why are there so many 404s for getting service items/shipments? Is that the response we get if there just aren't any service items/shipments? Here's a screenshot: 
Screen Shot 2021-05-21 at 12 36 00 PM

That's not particularly troubling if it's about lack of shipments/service items, but it would be kinda weird if it's because the move wasn't found. Either way, 404s are okay responses for load testing 🤷

duncan-truss:
> LGTM, thanks for fixing the important issues! If you do anything else, I would like renaming `data_list` to `data_dict`, but nothing else is required ☺️
> 
> I'm also curious, why are there so many 404s for getting service items/shipments? Is that the response we get if there just aren't any service items/shipments? Here's a screenshot:
> Screen Shot 2021-05-21 at 12 36 00 PM
> 
> That's not particularly troubling if it's about lack of shipments/service items, but it would be kinda weird if it's because the move wasn't found. Either way, 404s are okay responses for load testing 🤷

I agree `data_list` might be misleading, I was trying to keep things semi-coherent with the prime storage naming while introducing the new dict structure.  I'm not sure how I feel about adding types to the names but liked how you had it in the definition.

I'm not seeing the same failures you are, is your database filled from the devseed in the other PR https://github.com/transcom/mymove/pull/6669 ? 404s make sense if the moveId is doesn't exist or there are no service items or shipments like you said.

sandy-wright:
> I agree `data_list` might be misleading, I was trying to keep things semi-coherent with the prime storage naming while introducing the new dict structure. I'm not sure how I feel about adding types to the names (I guess this is a microcosm for when the underlying structure changes) but liked how you had it in the definition at least.
> 
> I'm not seeing the same failures you are, is your database filled from the devseed in the other PR [transcom/mymove#6669](https://github.com/transcom/mymove/pull/6669) ? 404s make sense if the moveId is doesn't exist or there are no service items or shipments like you said.

Oh yeah, the naming scheme in general is not great (my apologies!). If you can think of some other name besides `data_dict` that doesn't imply a data type but still gets the point across, please feel free to implement it! We can change the Prime version to match. 

I might not have devseeded in a bit, so that very well could have been the problem. That will probably occur a lot in experimental, since there is no devseed there ☺️

PR #73, opened 2021-05-27
[MB-8071] Load test SC and TOO edit move details ghc requests

duncan-truss:
## Description

This PR adds the first patch requests to update orders, allowances, and customer/service member records as a SC or TOO would on their respective Move Details page.

I added my first pytests for the OfficeDataStorageMixin that caught a bug with how I was storing some things.  I also added more local storage for duty station IDs because faker will not generate valid ones for us.

I had to add custom faker logic for the TAC field that has to be 4 characters long.  I also extended the parser to read the minimum and maximum definitions for fields like proGear weight and then call the faker random_int function with those range limits.

Here is the [corresponding PR in the mymove repo](https://github.com/transcom/mymove/pull/6696) that makes some changes to the ghc swagger to required backupContact fields and add the minimum and maximum validations that we already had on the model.

I added the `getCustomer` request because that is part of the header, which is displayed on the Move Details page as well.

Here is a downloadable report for simulating 500 office users with ~1000 moves in the database [report_1622145767.0901883.html.zip](https://github.com/transcom/milmove_load_testing/files/6556537/report_1622145767.0901883.html.zip)

## Reviewer Notes

I'm not sure I added the random_int params in the most generalized way, but I think it's a start and am happy to move things around if it proves useuful.

The 500 error I see for customer updater was when trying to update a backup contact for a service member who has no backup contact.  That might just be bad seed data but I can possibly add in a check for that case.

## Setup

```sh
make local_docker_build
make local_docker_office_up
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8071) for this change.

## Screenshots

Screen Shot 2021-05-27 at 4 02 57 PM
Screen Shot 2021-05-27 at 4 03 11 PM

### DIFF
--- a/tasks/office.py
+++ b/tasks/office.py
@@ -72,9 +72,11 @@ class OfficeDataStorageMixin:
         data_dict = self.local_store[object_key]
 
         if len(data_dict) >= self.DATA_LIST_MAX:
+            print("removing item from local store")
             num_to_delete = random.randint(1, len(data_dict) - 1)
+            print(f"num to delete {num_to_delete}")
             # Convert a dict to a list so we can take a slice by insertion order fifo
-            self.local_store[object_key] = dict(list(data_dict.items())[(num_to_delete):])
+            self.local_store[object_key] = dict(list(data_dict.items())[num_to_delete:])
 
         # Some creation endpoint auto-create multiple objects and return an array,
         # but each object in the array should still be considered individually here:
@@ -84,7 +86,7 @@ class OfficeDataStorageMixin:
             # merge the new data being added with the existing objects, not overwriting existing keys
             self.local_store[object_key] = {**normalized, **data_dict}
         else:
-            data_dict[object_data["id"]] = object_data
+            self.local_store[object_key][object_data["id"]] = object_data
 
 
 class ServicesCounselorTasks(OfficeDataStorageMixin, LoginTaskSet, ParserTaskMixin):
--- a/tasks/office.py
+++ b/tasks/office.py
@@ -3,11 +3,12 @@
 import logging
 import json
 import random
-from typing import Dict
+from typing import Dict, Set
 
 from locust import task, tag
 
 from utils.constants import (
+    CUSTOMER,
     MOVE,
     MOVE_TASK_ORDER,
     MTO_SERVICE_ITEM,
@@ -34,7 +35,12 @@ class OfficeDataStorageMixin:
     """
 
     DATA_LIST_MAX: int = 100
+    default_mto_ids: Dict[str, Set] = {
+        "destinationDutyStationID": set(),
+        "originDutyStationID": set(),
+    }
     local_store: Dict[str, Dict[str, Dict]] = {
+        CUSTOMER: {},
         MOVE_TASK_ORDER: {},
         MTO_SHIPMENT: {},
         MTO_SERVICE_ITEM: {},
@@ -72,9 +78,7 @@ class OfficeDataStorageMixin:
         data_dict = self.local_store[object_key]
 
         if len(data_dict) >= self.DATA_LIST_MAX:
-            print("removing item from local store")
             num_to_delete = random.randint(1, len(data_dict) - 1)
-            print(f"num to delete {num_to_delete}")
             # Convert a dict to a list so we can take a slice by insertion order fifo
             self.local_store[object_key] = dict(list(data_dict.items())[num_to_delete:])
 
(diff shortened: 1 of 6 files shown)


Ryan-Koch (approved):

utils/fake_data.py
```diff
@@ -120,24 +128,26 @@ def __init__(self):
...
     def get_random_choice(self, choices):
         """ Given a list of random choices, returns one of them. """
         return self.fake.random_element(choices)
 
-    def get_fake_data_for_type(self, data_type):
+    def get_fake_data_for_type(self, data_type, params=None):
         """ Given a specific data type, returns faker data for that type (if a mapping exists). """
         try:
-            return self.data_types[data_type]()
+            return self.data_types[data_type](*params)
```
sandy-wright, line 150: nice, this is a good addition. `**params` instead might be a little safer, but I think this is probably fine

utils/fields.py
```diff
@@ -16,6 +16,7 @@ class BaseAPIField:
     name: str = ""
     required: bool = False
     discriminator_values: list = None
+    definition: dict = None
```
sandy-wright, line 19: HmmmMMM, I would pick apart the definition here. The idea is to be independent from the YAML at this level. I would try to be explicit instead:
```python
minimum: int = None  # unless we want default values here
maximum: int = None
```
And then something like this in `generate_fake_data`:
```python
   params = [
       minimum if minimum else 0,
       maximum if maximum else 0,  # what should this else be? I don't remember how randint handles no max
   ]
   
    return faker.get_fake_data_for_type(self.data_type, params if params else [])
```

Then you would need to change the `BaseAPIField` initialization to something like:
```python
return BaseAPIField(data_type=data_type, name=name, minimum=definition.get("minimum") ... etc
```

You could also subclass BaseAPIField into `IntegerField` or something as well. But the idea is to be definition-agnostic in the fields level.

utils/parsers.py
```diff
@@ -327,7 +327,7 @@ def _parse_typed_field(self, name, field_type, field_format):
             data_type = self._approximate_str_type(name)
 
         if data_type:
-            return BaseAPIField(data_type=data_type, name=name)
+            return BaseAPIField(data_type=data_type, name=name, definition=definition)
```
sandy-wright, line 330: As I mentioned above, I would explicitly pass in minimum and maximum here instead of the whole definition.

utils/fields.py
```diff
@@ -60,7 +61,13 @@ def generate_fake_data(self, faker, **kwargs):
...
             return overrides[self.name]
 
-        return faker.get_fake_data_for_type(self.data_type)
+        params = (
+            [self.definition.get("minimum", 0), self.definition.get("maximum", 9999)]
+            if self.data_type == DataType.INTEGER
+            else []
+        )
+
+        return faker.get_fake_data_for_type(self.data_type, params)
```
sandy-wright, line 70: Would you be willing to add a test for the min/max situation to `test_fields` and/or `test_parsers`? Optional ☺️

sandy-wright (commented):
Aha, I'm a little late! But I'll submit these comments anyway so I can remember them at least. Thanks for adding those tests!

duncan-truss:
> Aha, I'm a little late! But I'll submit these comments anyway so I can remember them at least. Thanks for adding those tests!

Sorry there wasn't more time to revise and discuss this PR.  This is exactly the sort of feedback I was looking for thank you!

tasks/office.py
```diff
@@ -363,6 +504,102 @@ def get_service_items(self, overrides=None):
...
+    def get_customer(self, overrides=None):
+        """
+        Fetches a single service member's orders
+        """
+        # If id was provided, get that specific one. Else get any stored one.
+        object_id = overrides.get("id") if overrides else None
+        order = self.get_stored(ORDER, object_id)
+        if order is None:
+            logger.info("Skipping get customer no order is stored yet")
+            return
```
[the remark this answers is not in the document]
duncan-truss: I like that idea, I see that pattern emerging now in multiple places.  The skip tends to only happen before the calls to the queue endpoints fill the local store with data.  I was trying to not hard code that setup sequence but that is another option to explore.

utils/fake_data.py
duncan-truss: So **params would handle the case where we wanted a Dict's items converted to params?  I think that's what I understood is the difference from the docs.

utils/fields.py
duncan-truss: Yea this is primarily what I needed your eyes on, I don't know the parser inside and out so this seems like the better abstraction level.

PR #75, opened 2021-09-29
Use listMoves endpoint instead of fetchMTOUpdates

sandy-wright:
## Description

Removes all references to `fetchMTOUpdates` and leverages the `listMoves` endpoint instead. Note that there are a number of _other_ issues with the Prime workflow at the moment that might be related to input. This PR doesn't address those.

To run the Prime workflow, first run your server in `mymove`:

```sh
make server_run
```

Then, in `milmove_load_testing`, use the `prime_workflow.py` locust file:

```sh
make load_test_prime_workflow
```

### DIFF
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -215,15 +215,15 @@ class PrimeTasks(PrimeDataStorageMixin, ParserTaskMixin, CertTaskMixin, TaskSet)
     tags where appropriate to make filtering for custom tests easier.
     """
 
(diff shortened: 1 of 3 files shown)


gidjin (approved):
No failures in the move endpoint! I did see failures for `/prime/v1/move-task-orders/{moveTaskOrderID}/post-counseling-info` endpoint but that feels unrelated to this fix.

PR #83, opened 2021-11-09
[MB-10361] Use custom parameters to configure User Class Weights

ahobson:
## Description

This is an example of how we could use custom weights to allow us to control which tests are run. Setting a user weight to 0 would remove them.

I tested this locally with locust 2.5.0 and it worked for me. I could run a test and then stop it and run with another set of user weights and it worked.

### DIFF
--- a/locustfiles/all.py
+++ b/locustfiles/all.py
@@ -1,0 +1,136 @@
+# -*- coding: utf-8 -*-
+""" Locust test for all APIs """
+from locust import HttpUser, between, events
+
+from utils.hosts import MilMoveHostMixin, MilMoveDomain, clean_milmove_host_users
+from utils.parsers import GHCAPIParser, PrimeAPIParser, SupportAPIParser
+from tasks import PrimeTasks, SupportTasks, ServicesCounselorTasks, TOOTasks
+
+# from flask import make_response, redirect
+from gevent.pool import Group
+
+
+# init these classes just once because we don't need to parse the API over and over:
+prime_api = PrimeAPIParser()
+support_api = SupportAPIParser()
+ghc_api = GHCAPIParser()
+
+
+class PrimeUser(MilMoveHostMixin, HttpUser):
+    """
+    Tests the Prime API.
+    """
+
+    # These attributes are used in MilMoveHostMixin to set up the proper hostname for any MilMove environment:
+    local_port = "9443"
+    domain = MilMoveDomain.PRIME  # the base domain for the host
+    is_api = True  # if True, uses the api base domain in deployed environments
+
+    # This attribute is used for generating fake requests when hitting the Prime API:
+    parser = prime_api
+
+    # These are locust HttpUser attributes that help define and shape the load test:
+    wait_time = between(0.25, 9)  # the time period to wait in between tasks (in seconds, accepts decimals and 0)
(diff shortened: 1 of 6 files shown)


locustfiles/all.py
```diff
@@ -0,0 +1,148 @@
...
+            return render_template("get_user_classes.html", **render_ctx)
+
+        @extend.route("/set_user_classes", methods=["POST"])
+        def set_user_classes():
+            assert request.method == "POST"
+
+            class_options = request.form.getlist("user_classes_option")
+            new_classes = [original_user_classes[n] for n in class_options]
+            setattr(environment, "user_classes", new_classes)
+            environment._remove_user_classes_with_weight_zero()
```
duncan-truss, line 139: This errors out for me, because it's an internal method I'm guessing? Since you're setting the user_classes above to what was selected we wouldn't need to set the unselected classes to a weight of zero I'm assuming.

duncan-truss:
One thing I noticed is the stop test button has vanished for me after changing the user classes and status shows stopped once selecting new user classes that are running properly 

Screen Shot 2021-11-10 at 3 08 20 PM

duncan-truss:
I think there is a way to add custom arguments to appear on the initial screen but I have a feeling that is in version 2 and we are still on Locust version 1.  Have there been any discussions on upgrading by chance?  On first glance I don't see major breaking changes (famous last words).

https://github.com/locustio/locust/pull/1881
https://github.com/locustio/locust/blob/master/examples/add_command_line_argument.py
https://github.com/locustio/locust/releases/tag/2.0.0

locustfiles/all.py
ahobson: I *think* this might be in a different version of locust. It doesn't error out for me, so hopefully the pipenv changes will make it work for you, too

ahobson:
> I think there is a way to add custom arguments to appear on the initial screen but I have a feeling that is in version 2 and we are still on Locust version 1. Have there been any discussions on upgrading by chance? On first glance I don't see major breaking changes (famous last words).
> 
> [locustio/locust#1881](https://github.com/locustio/locust/pull/1881) https://github.com/locustio/locust/blob/master/examples/add_command_line_argument.py https://github.com/locustio/locust/releases/tag/2.0.0

THAT IS IT! THANK YOU

ahobson:
> One thing I noticed is the stop test button has vanished for me after changing the user classes and status shows stopped after selecting new user classes and them starting up properly.
> 
> Screen Shot 2021-11-10 at 3 08 20 PM

I think this new version fixes that problem. At least, it does for me

felipe-lee:
@ahobson, we've pulled this story into our current sprint, but I've just done a diff with master and I think at this point all the relevant changes might have been merged in other branches. Am I correct in that? Or am I missing something?

ahobson:
> @ahobson, we've pulled this story into our current sprint, but I've just done a diff with master and I think at this point all the relevant changes might have been merged in other branches. Am I correct in that? Or am I missing something?

@felipe-lee You are right! I forgot we had this as a separate story and I accidentally sneaked it into another PR. Sorry about that

felipe-lee:
Awesome, no problem. I'll close this and the ticket out then!

PR #84, opened 2021-11-15
MB-10413: Convert to using pipenv to manage dependencies

ahobson:
## Description

Pipenv lets us manage dependencies and creates a lock file so we can have reproducible builds.

In addition, `direnv` supports it natively and will install the dependencies and manage the virtualenv for us.

I also took the opportunity to get the pre-commit in sync with the dependencies and use a `pyproject.toml`

Finally, I fixed the tests and enabled them in CI

## Reviewer Notes

I've tested this out with nix and as much as I can with brew, although I don't use it extensively, so any validation you all can do would be great

## Setup

Check out the code and try to follow the README to get your environment set up

PR #85, opened 2021-11-17
MB-10438: Update task definition on every build

ahobson:
## Description

* Push a unique image tag per build based on the git sha
* Create a new task definition for every deploy with that image tag
* Override the LOCUSTFILE as necessary

## Reviewer Notes

I tested this out with a build override of this branch and it worked well

I based this on the [pipenv work](#84) since the Dockerfile needed to be updated. That means it's really only the last commit that's relevant

### DIFF
--- a/locustfiles/all.py
+++ b/locustfiles/all.py
@@ -6,7 +6,6 @@ from utils.hosts import MilMoveHostMixin, MilMoveDomain, clean_milmove_host_user
 from utils.parsers import GHCAPIParser, PrimeAPIParser, SupportAPIParser
 from tasks import PrimeTasks, SupportTasks, ServicesCounselorTasks, TOOTasks
 
-# from flask import make_response, redirect
 from gevent.pool import Group
 
 
--- a/locustfiles/all.py
+++ b/locustfiles/all.py
@@ -1,83 +1,32 @@
 # -*- coding: utf-8 -*-
 """ Locust test for all APIs """
-from locust import HttpUser, between, events
+from locust import events
 
-from utils.hosts import MilMoveHostMixin, MilMoveDomain, clean_milmove_host_users
+from utils.hosts import clean_milmove_host_users
 from utils.parsers import GHCAPIParser, PrimeAPIParser, SupportAPIParser
-from tasks import PrimeTasks, SupportTasks, ServicesCounselorTasks, TOOTasks
+
+from prime import PrimeUser, SupportUser
+from prime_workflow import PrimeWorkflowUser
+from office import ServicesCounselorUser, TOOUser
 
 from gevent.pool import Group
+
+########################################################################
+#
+#  TODO: This file contains all the users in one place. If a user is
+#  created in another locustfile, it should be added here along with a
+#  custom argument for controlling the weight
+#
+#  Maybe we should just consolidate all user definitions into this
+#  file, or maybe we should move the user definitions elsewhere?
+#
+########################################################################
 
 
 # init these classes just once because we don't need to parse the API over and over:
 prime_api = PrimeAPIParser()
 support_api = SupportAPIParser()
 ghc_api = GHCAPIParser()
-
-
-class PrimeUser(MilMoveHostMixin, HttpUser):
-    """
-    Tests the Prime API.
-    """
-
-    # These attributes are used in MilMoveHostMixin to set up the proper hostname for any MilMove environment:
-    local_port = "9443"
-    domain = MilMoveDomain.PRIME  # the base domain for the host
-    is_api = True  # if True, uses the api base domain in deployed environments
-
-    # This attribute is used for generating fake requests when hitting the Prime API:
-    parser = prime_api
-
-    # These are locust HttpUser attributes that help define and shape the load test:
-    wait_time = between(0.25, 9)  # the time period to wait in between tasks (in seconds, accepts decimals and 0)
-    tasks = {PrimeTasks: 1}  # the set of tasks to be executed and their relative weight
-
-
-class SupportUser(MilMoveHostMixin, HttpUser):
-    """
-    Tests the Support API.
-    """
-
-    local_port = "9443"
-    domain = MilMoveDomain.PRIME
-    is_api = True
-
-    parser = support_api
-
-    wait_time = between(0.25, 9)
-    tasks = {SupportTasks: 1}
-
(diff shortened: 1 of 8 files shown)


scripts/codebuild
```diff
@@ -6,30 +6,67 @@ set -x
 echo "${DOCKER_PASSWORD}" | \
   docker login --username "${DOCKER_USERNAME}" --password-stdin
 
-#LOCUSTFILE can be set in the build overrides of the CodeBuild config
-docker build -t "${ECR_REPO}:exp" --build-arg locustfile="${LOCUSTFILE}" .
+tag=${CODEBUILD_RESOLVED_SOURCE_VERSION:-exp}
```
duncan-truss, line 9: The Codebuild docs say this would be the git commit hash I think but I can't match up the value to a commit in this PR.

Screen Shot 2021-11-22 at 10 39 30 AM

https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html

duncan-truss: Whoops I see it now GitHub just had difficulty searching for it maybe because of the force push https://github.com/transcom/milmove_load_testing/commit/f6af768fea72d92ea6e95b24ab9b057d1c44213e

scripts/codebuild
```diff
@@ -6,30 +6,67 @@ set -x
...
+                          --task-definition "${task_definition_name}" | \
+                        jq "${jq_filter}")
+
+# LOCUSTFILE can be set in the build overrides of the CodeBuild config
+: "${LOCUSTFILE:=/app/locustfiles/prime.py}"
+# update with new settings (e.g. image)
+new_task_definition=$(
+  echo "${raw_task_definition}" | \
+    jq ".containerDefinitions[0].image = \"${image}\"" | \
+    jq ".containerDefinitions[0].entryPoint = [\"locust\"]" | \
```
duncan-truss, line 56: I think this is already the default entrypoint in the Dockerfile. I know I only overrode it in the terraform because it needed to evaluate the env var.

scripts/codebuild
```diff
@@ -6,30 +6,67 @@ set -x
...
+joinedKeys=${joinedKeys:1}
+
+jq_filter=".taskDefinition | {${joinedKeys}}"
+
+raw_task_definition=$(aws ecs describe-task-definition \
+                          --task-definition "${task_definition_name}" | \
+                        jq "${jq_filter}")
+
+# LOCUSTFILE can be set in the build overrides of the CodeBuild config
+: "${LOCUSTFILE:=/app/locustfiles/prime.py}"
```
duncan-truss, line 51: I had to look up what `:` means for shell scripting, I guess it's just a noop value of true?

duncan-truss (approved):
Looking at the Codebuild logs and the ECS service looks like it deployed successfully
Screen Shot 2021-11-22 at 10 57 55 AM

scripts/codebuild
ahobson: Sorry for confusing things. I guess I should add a comment explaining `CODEBUILD_RESOLVED_SOURCE_VERSION`

scripts/codebuild
ahobson: It is, but I was trying to be explicit in case it was overridden elsewhere.

I have started another branch that uses python to build the container definition like we do in go for MilMove. I think it's much clearer, but it's not ready yet. When I have time ...

scripts/codebuild
ahobson: Sorry, yeah, this is a convention I've seen elsewhere. When using `:=` variable substitution, that statement doesn't do anything but modify the variable, so using `:` is the way to indicate that so shellcheck doesn't complain.

Is there a comment or something I should add?

scripts/codebuild
ahobson: I just added a comment so hope that helps

scripts/codebuild
ahobson: I added a comment here, too

locustfiles/all.py
```diff
@@ -0,0 +1,136 @@
+# -*- coding: utf-8 -*-
+""" Locust test for all APIs """
+from locust import HttpUser, between, events
+
+from utils.hosts import MilMoveHostMixin, MilMoveDomain, clean_milmove_host_users
+from utils.parsers import GHCAPIParser, PrimeAPIParser, SupportAPIParser
+from tasks import PrimeTasks, SupportTasks, ServicesCounselorTasks, TOOTasks
+
+# from flask import make_response, redirect
```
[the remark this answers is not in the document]
ahobson: Not used anymore. Thank you, removed!

locustfiles/all.py
```diff
@@ -0,0 +1,136 @@
...
+
+    local_protocol = "http"
+    local_port = "8080"
+    domain = MilMoveDomain.OFFICE
+
+    # This attribute is used for generating fake requests when hitting the GHC API:
+    parser = ghc_api
+
+    wait_time = between(0.25, 9)
+    tasks = {TOOTasks: 1}
```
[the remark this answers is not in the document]
ahobson: That's my vote. We can have one fall to run all the tests and use the UI to disable the users if we don't want to run any particular user.

I guess this is kind of an experiment so I'd be ok with leaving the duplication for now, but if you all would prefer to go ahead and consolidate, I'm fine with that, too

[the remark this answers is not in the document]
felipe-lee: I'm fine with leaving the duplication since we're experimenting, but can you add `TODO`'s for the ones that we should revisit deleting? Essentially I just want to make sure that a future person that sees both sees something in the codebase indicating we knowingly put in the duplication and that the plan is to likely delete after testing it for a while.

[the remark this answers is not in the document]
felipe-lee: I'm fine with leaving the duplication since we're experimenting, but can you add `TODO`'s for the ones that we should revisit deleting? Essentially I just want to make sure that a future person that sees both sees something in the codebase indicating we knowingly put in the duplication and that the plan is to likely delete after testing it for a while.

felipe-lee (approved):
LGTM

locustfiles/all.py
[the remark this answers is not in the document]
felipe-lee: Awesome, thank you!

PR #86, opened 2021-11-23
MB-10413 Explicitly set python_full_version in `Pipfile`

felipe-lee:
## Description

While prepping for demo, I realized that my venv was using python 3.9.9. Apparently we need to explicitly set `python_full_version` in `Pipfile`. Otherwise defaults to system default, as long as it's 3.9.*. So for me, my venv was being created with python 3.9.9. I guess I must have updated my global python version since I last messed with this just last week haha.

This then led to an issue with the docker image we were using since it didn't have python 3.9.6 nor pyenv/asdf to install the right python version. Switched to using CircleCI's python docker image since it has what we need (even comes with `pipenv`) and [CircleCI built it specifically for running in CI](https://circleci.com/docs/2.0/circleci-images/#next-generation-convenience-images). 

The problem with the python image though, is that it doesn't have `pre-commit` like our previous docker image did. Thankfully since this project is python-based and `pre-commit` is too, we can install it with `pip` so I added it to the `Pipfile` dependencies. It also makes it so that we are using the same `pre-commit` version in CircleCI, nix, and non-nix installs.

There were also a few other updates to the dependencies that pipenv found.

## Reviewer Notes

Can someone set up a global python version that is 3.9.*, but not 3.9.6 and make sure it also sets up the proper python version for their venv?

## Setup

Follow setup instructions. If you already have the repo set up, you can run

```bash
pipenv --rm
```

to clear out your current venv.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10413) for this change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,7 @@
 #! /usr/bin/make
 
-PRE_COMMIT:=/usr/local/bin/pre-commit
-
 ifdef NIX_PROFILE
 	USE_NIX:=true
-	PRE_COMMIT:=$(NIX_PROFILE)/bin/pre-commit
 endif
 
 PRIME_LOCUSTFILES=/app/locustfiles/prime.py
@@ -24,9 +21,8 @@ install_python_deps: ## Install all python dependencies/requirements
 
 .PHONY: ensure_pre_commit
 ensure_pre_commit: .git/hooks/pre-commit  ## Ensure pre-commit is installed
-.git/hooks/pre-commit: $(PRE_COMMIT)
-	pre-commit install
-	pre-commit install-hooks
+.git/hooks/pre-commit: $(shell which pre-commit)
+	pre-commit install --install-hooks
 
 .PHONY: pre_commit_update_deps
 pre_commit_update_deps:  ## Update pre-commit dependencies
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -24,9 +21,8 @@ install_python_deps: ## Install all python dependencies/requirements
 
 .PHONY: ensure_pre_commit
 ensure_pre_commit: .git/hooks/pre-commit  ## Ensure pre-commit is installed
-.git/hooks/pre-commit: $(PRE_COMMIT)
-	pre-commit install
-	pre-commit install-hooks
+.git/hooks/pre-commit: $(shell which pre-commit)
```
felipe-lee, line 24: Changed to using the shell to get the path because it's easier to maintain. Otherwise it would have changed again to account for installing via `pip` instead of `brew`

duncan-truss: It looks like `make setup` calls `pipenv install`, which installs `pre-commit`.  Do we still need this target then?

duncan-truss (approved):
Removing my pipenv and repeating install instructions seemed to work, including running tests and pre-commit hooks 🎉

Makefile
felipe-lee: `pipenv` installs the pre-commit package, but this command installs the hooks that pre-commit will run.

PR #88, opened 2021-12-01
Update python dependency versions

felipe-lee:
## Description

Just happened to notice some packages had updates and since we don't have a process in place, I figure a quick update was good.

## Setup

I was just going to let CI run the tests, especially since some load tests are failing as is. If you'd like to run the repo tests locally, you can run

```sh
pipenv sync -d
```

then

```sh
pytest
```

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -92,11 +92,11 @@
         },
         "charset-normalizer": {
             "hashes": [
-                "sha256:e019de665e2bcf9c2b64e2e5aa025fa991da8720daa3c1138cadd2fd1856aed0",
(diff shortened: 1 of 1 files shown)


pearl-truss:
Thanks for this update! I think this raises a really good point about how do we want to handle dependency updates for this repo...? We may or may not need to answer this yet but whatever we decide should be documented somewhere

pearl-truss (approved):

felipe-lee:
> Thanks for this update! I think this raises a really good point about how do we want to handle dependency updates for this repo...? We may or may not need to answer this yet but whatever we decide should be documented somewhere

Yeah we definitely need a plan. I created a ticket to come up with one: https://ustcdp3.slack.com/archives/CP6F568DC/p1638401546189200

PR #74, opened 2021-09-02
MB-9638: Add ecs deploy + codebuild config + nix updates

ahobson:
## Description

Provide a way to deploy to ECS

## Reviewer Notes

Is there anything you would like reviewers to give additional scrutiny?

## Setup

Add any steps or code to run in this section to help others prepare to run your code:

```sh
echo "Code goes here"
```

## Code Review Verification Steps

* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](tbd) for this change.
* [This article](tbd) explains more about the approach used.

PR #89, opened 2021-12-02
MB-10464 Add reasons for why local docker setup isn't supported anymore

felipe-lee:
## Description

In a previous PR (#74) the local docker setup broke and in a later PR (#78) the instructions were updated indicating the local docker setup wouldn't be supported anymore, but didn't give reasoning. This PR adds some reasons for the change so that future maintainers can have some context.

## Reviewer Notes

Given that the local docker setup is broken, would it maybe be better to delete the instructions and have my git commit list the reasons?

## Setup

Easiest way to read the formatted README would probably be to look at the [README on this branch in github](https://github.com/transcom/milmove_load_testing/blob/MB-10464-update-docs-to-indicate-why-docker-is-no-longer-supported-locally/README.md).

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10464) for this change.


README.md
```diff
@@ -182,8 +182,17 @@ entirely at a future date.
 
 #### Alternative Setup: Docker
 
-It is also possible to run load tests from within a Docker container, eliminating the need to set up a valid python
-environment. This requires Docker and `docker-compose` to be installed on your machine. Get them here:
+It might be possible to run load tests from within a docker container, though given recent updates this seems to not be
```
pearl-truss, line 185: Nit: maybe consider dropping "recent" since some time might pass before this is looked at again. Also, it might be worth linking to the PR's that introduced the breaking changes here

pearl-truss:
> Given that the local docker setup is broken, would it maybe be better to delete the instructions and have my git commit list the reasons?

I'm in favor of this since it's not working I don't think we need instructions for it anymore but it would be nice to have a reference to the reasons

ronaktruss:
I'm also in favor of removing the instructions and indicating the reasons that we don't support it. If we do this we should also remove some of the code/files that are specific for this functionality such as `local_locust.py` and `docker-compose.local.yaml`.

felipe-lee:
Sounds good, I can make those updates, thanks for the feedback!

felipe-lee:
@pearl-truss @ronaktruss , I deleted the local docker setup entirely now, added the PR links to the README, and modified the details a little.

I'd mentioned in my description that maybe we should just remove everything from the README regarding local docker `asdf`, but now that I think about it, having the "Unsupported Setup" section might be good to keep so that as we hand this off to other teams. It's more visible why we don't have more setup options rather than having to dig through the commit history 🤔

PR #91, opened 2021-12-07
[MB-10531] Fix failing prime and office load tests

duncan-truss:
## Description

This story requires 2 PRs, this one and [one in the mymove repo](https://github.com/transcom/mymove/pull/7862), to address common failures when running the prime and office load tests that make it difficult to know if they are working as expected.

Prime tests that were failing:
- The prime creates it's own move task orders and there were a lot of eTag failures when it came time to update the move for say post counseling information.  It turns out there was a bug in our mymove code that returned a move available to the prime with the outdated eTag because of a pointer issue not updating the returned move.

Office tests that were failing:
- Both the Services Counselor and the TOO can update orders and allowances in the office app.  Managing one endpoint for these different use cases become difficult so they were split out but load tests were not updated so the Services Counselor requests were wrong and returning unauthorized.
- In the load tests themselves we were saving customers to the shared data store but depending on the endpoint not all associations were fully loaded in the response from the API.  I removed a call to add customers unless they are fetched from getCustomer, which saves a full record.
- There are cases of a non complete Service Member being created without a residential address and backup contacts in the seed data for office moves.  This will never be the case in reality as those are required fields during customer onboarding, this is just a case of inconsistent seed data setup.
- When calling update orders as the TOO we check to see if the move can be set back to Approved in the case of amended orders being acknowledged and other requirements being true.  If the TOO tries to operate on a move in services counseling this would result in error if trying to change the status to approved.  This is a limitation of the current shared data store, so I made a check to not use moves in that status if it's the TOO user and not the Services Counselor.

Most of the errors should now be expected 412 eTag errors from 2 or more operations working on the same move prior to it being updated in the data store.

I will update the ghc api URL to point back to master once I merge in the other PR before merging this one.

**Prime tests still failing:**
- update shipment Geocode errors for addresses
- updating shipment status (diversion, cancellation) from not allowed status
- creating payment request when shipment is missing actual weight
- create mto shipment with 2 of the same agent types

## Reviewer Notes

## Setup

Be sure to checkout the PR branch in the mymove repo
```sh
make db_dev_e2e_populate
make server_run
```

In the load testing repo
```sh
make load_test_prime
```

```sh
make load_test_office
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10531 for this change.


utils/parsers.py
```diff
@@ -480,4 +480,6 @@ def _custom_request_validation(self, path, method, request_data):
 class GHCAPIParser(APIParser):
     """Parser class for the GHC API."""
 
-    api_file = "https://raw.githubusercontent.com/transcom/mymove/master/swagger/ghc.yaml"
+    api_file = (
+        "https://raw.githubusercontent.com/transcom/mymove/0d619037ee393bdba7ea9afa93b5a18529c2d904/swagger/ghc.yaml"
```
pearl-truss, line 484: @duncan-truss is this a code change that'll be reverted when the mymove PR lands?

duncan-truss: Yes, hopefully I'll remember to switch it back once this mymove PR merges in first

felipe-lee (approved):
LGTM, I posted a more detailed review in the mymove PR, but it applies to both really

PR #92, opened 2021-12-08
MB-10536 Add starting point for customer seed data for prime

ronaktruss:
## Description

This adds a starting point for initializing seed data within the `prime.py` locust test flow. For each user we are logging into dev-local and attempting to go through the entire customer flow. Right now these changes just create the user and adds all of the information needed to setup their profile.

Note: Some of this code should be refactored if it is being used in other tests. 

## Reviewer Notes

## Setup

```sh
locust -f locustfiles/prime.py --host local
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10536) for this change.

### DIFF
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -267,7 +267,6 @@ class PrimeTasks(PrimeDataStorageMixin, ParserTaskMixin, CertTaskMixin, TaskSet)
         payload = self.fake_request(
             "/service_members/{serviceMemberId}/backup_contacts", "post", INTERNAL_API_KEY, overrides
         )
-        print(payload)
         self.client.post(
             self.customer_path(f"/internal/service_members/{service_member_id}/backup_contacts"),
             name="/internal/service_members/{serviceMemberId}/backup_contacts",
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -9,6 +9,7 @@ from typing import Dict
 from locust import tag, task, TaskSet
 
 from utils.constants import (
+    INTERNAL_API_KEY,
     TEST_PDF,
     ZERO_UUID,
     PRIME_API_KEY,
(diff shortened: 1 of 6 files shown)


pearl-truss (commented):
Thanks for pushing up these changes! Can you call out which code would need to be refactored to be used in other test? Also which other tests?

utils/constants.py
```diff
@@ -37,7 +38,9 @@ class DataType(ValueEnum):
     """Swagger data types that we expect to deal with. Uses camelcase in values to match."""
```
pearl-truss, line 38: Nit: I think this comment needs updating since not all the fields are camelCase anymore

pearl-truss (approved):
I left some minor comments but I think the code changes look good!

tasks/prime.py
```diff
@@ -215,6 +216,65 @@ class PrimeTasks(PrimeDataStorageMixin, ParserTaskMixin, CertTaskMixin, TaskSet)
...
 
+    def __init__(self, parent):
+        self.csrf_token = None
+        self.session_token = None
+        super().__init__(parent)
+
+    def customer_path(self, url: str) -> str:
+        return f"{self.user.alternative_host}{url}"
+
+    def on_start(self):
```
ronaktruss, line 227: @pearl-truss It's likely that the entirety of the on_start function should be refactored. The first part of the code is just setting up the cookies so that we can login while the latter portion is going through the APIs to do the profile setup. The cookie setup functionality already exist in an abstracted form in the `LoginTaskSet` class. I'm hoping that once we have a crude version of this working and the tests are fixed, we will more context to finalize how do we structure the code.

utils/constants.py
ronaktruss: Updated

felipe-lee (approved):
LGTM, but I do agree we should refactor in future iterations as we learn which bits would be useful in other areas. It'd also be good to refactor going forward to make the logic easier to read since right now it's doing a series of requests so the `on_start` method is pretty long.

tasks/prime.py
```diff
@@ -215,6 +216,65 @@ class PrimeTasks(PrimeDataStorageMixin, ParserTaskMixin, CertTaskMixin, TaskSet)
...
+        if resp.status_code != 200:
+            self.interrupt()
+
+        logged_in_user = self.client.get(self.customer_path("/internal/users/logged_in"))
+        json_resp = logged_in_user.json()
+        service_member_id = json_resp["service_member"]["id"]
+        email = json_resp["email"]
+        user_id = json_resp["id"]
+
+        origin_duty_stations = self.client.get(self.customer_path("/internal/duty_stations?search=29"))
```
duncan-truss, line 245: For the prime it won't matter what origin duty station is selected but it might be good to document which one is actually being selected first in the list when it comes to the office GBLOC and whether the move goes to services counseling or not.

tasks/prime.py
```diff
@@ -215,6 +216,65 @@ class PrimeTasks(PrimeDataStorageMixin, ParserTaskMixin, CertTaskMixin, TaskSet)
...
+        overrides = {
+            "id": service_member_id,
+            "user_id": user_id,
+            "edipi": "9999999999",
+            "personal_email": email,
+            "email_is_preferred": True,
+            "current_station_id": current_station_id,
+        }
+
+        payload = self.fake_request("/service_members/{serviceMemberId}", "patch", INTERNAL_API_KEY, overrides, True)
```
duncan-truss, line 257: This has bit me before in the seed data, if Marine Corps gets chosen for the affiliation that will send it to a different office queue regardless of origin duty station.  That won't affect the prime but just be aware some of those moves will get routed differently and folks may wonder why it disappeared.

utils/constants.py
```diff
@@ -34,10 +35,14 @@
...
 class DataType(ValueEnum):
-    """Swagger data types that we expect to deal with. Uses camelcase in values to match."""
+    """
+    Swagger data types that we expect to deal with. The latest pattern uses camelCase, while the older pattern uses snake_case.
+    """
 
     FIRST_NAME = "firstName"
+    FIRST_NAME_VARIANT = "first_name"  # matches older swagger format
     LAST_NAME = "lastName"
+    LAST_NAME_VARIANT = "last_name"  # matches older swagger format
```
duncan-truss, line 45: I know James added middle name and suffix to the fake data spreadsheet probably since this was last touched if we want to take advantage of those down the line.

tasks/prime.py
duncan-truss: This runs for every instance of the prime user too right?  So if we generate 100 prime users it would initiate 100 new customers?

duncan-truss (approved):
This is exciting, thanks for getting this started.

tasks/prime.py
ronaktruss: The duty station chose is `MCAGCC Twentynine Palms`

tasks/prime.py
ronaktruss: Thanks for that info! I also see that some of the duty stations have an affiliation associated with them. What is this used for?

tasks/prime.py
ronaktruss: Yeah, that is correct. There are ways to run a function per test rather than per user, but we don't have access to the client that makes the API calls there (at least by default).

PR #93, opened 2021-12-14
MB-10652 create shipment

ronaktruss:
## Description

This fixes tests for creating an MTO shipment when multiple agents have the same type. This also adds a tests for an intentional failure, for having multiple agents of the same type.

image

## Reviewer Notes

Do we want to maintain logging for expected failure cases? Using the `check_response` helper method causes the output in the terminal to be verbose. When we have an error check_response logs the entire json payload which in this case is ~45 lines.

## Setup

```sh
make load_test_prime
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10652) for this change.

### DIFF
--- a/tasks/base.py
+++ b/tasks/base.py
@@ -10,7 +10,7 @@ from requests import Response
 logger = logging.getLogger(__name__)
 
 
-def check_response(response: Response, task_name="Task", request=None, expected_status_code: str = None):
+def check_response(response: Response, task_name="Task", request=None, expected_status_code: str = "2"):
     """
     Logs the status code from the response and converts it from JSON into a python dictionary we can work with. If the
     status code doesn't match (by default 2xx), it can also log any request data that was sent in for the sake of debugging.
@@ -29,9 +29,6 @@ def check_response(response: Response, task_name="Task", request=None, expected_
     except (json.JSONDecodeError, TypeError):
         logger.exception("Non-JSON response.")
         return None, False
-
-    if expected_status_code is None:
-        expected_status_code = "2"
 
     if not str(response.status_code).startswith(expected_status_code):
         logger.error(f"⚠️ {task_name} failed.\n{json.dumps(json_response, indent=4)}")
--- a/tasks/base.py
+++ b/tasks/base.py
@@ -10,15 +10,16 @@ from requests import Response
 logger = logging.getLogger(__name__)
 
 
-def check_response(response: Response, task_name="Task", request=None):
+def check_response(response: Response, task_name="Task", request=None, expected_status_code: str = None):
     """
     Logs the status code from the response and converts it from JSON into a python dictionary we can work with. If the
-    status code wasn't a success (2xx), it can also log any request data that was sent in for the sake of debugging.
+    status code doesn't match (by default 2xx), it can also log any request data that was sent in for the sake of debugging.
     Returns the dictionary representation of the response content and a boolean indicating success or failure.
 
     :param response: HTTP Response class from the Python requests framework
     :param task_name: str, optional name of the tasks
     :param request: any type, optional data to print for debugging a failed response
+    :param expected_status_code: str, expected code of response if no value is provided expects 2xx
     :return: tuple(dict, bool)
     """
     logger.info(f"ℹ️ {task_name} status code: {response.status_code} {response.reason}")
@@ -29,7 +30,10 @@ def check_response(response: Response, task_name="Task", request=None):
         logger.exception("Non-JSON response.")
         return None, False
 
-    if not str(response.status_code).startswith("2"):
+    if expected_status_code is None:
+        expected_status_code = "2"
+
+    if not str(response.status_code).startswith(expected_status_code):
         logger.error(f"⚠️ {task_name} failed.\n{json.dumps(json_response, indent=4)}")
         if request:
             try:
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -544,8 +544,11 @@ class PrimeTasks(PrimeDataStorageMixin, ParserTaskMixin, CertTaskMixin, TaskSet)
         if mto_shipment.get("agents") is None:
             return  # can't update agents if there aren't any
 
+        overrides = {}
+        mto_agents = mto_shipment["agents"]
         mto_agent = mto_shipment["agents"][0]
-        overrides = {"agentType": mto_agent["agentType"]}  # updates the agent, but keeps the same agent type
+        if len(mto_agents) >= 2:
+            overrides = {"agentType": mto_agent["agentType"]}  # ensure agentType does not change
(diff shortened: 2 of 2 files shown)


tasks/prime.py
```diff
@@ -362,6 +369,48 @@ def create_mto_shipment(self, overrides=None):
...
+
+        headers = {"content-type": "application/json"}
+        resp = self.client.post(
+            prime_path("/mto-shipments"),
+            name=prime_path("/mto-shipments — expected failure"),
+            data=json.dumps(payload),
+            headers=headers,
+            **self.user.cert_kwargs,
+        )
+        check_response(resp, "createMTOShipment", payload)
```
ronaktruss, line 412: As mentioned in the PR. Do we need this? We are generally using this throughout the tests but this may not be that useful for intentionally failed tests.

tasks/prime.py
```diff
@@ -362,6 +369,48 @@ def create_mto_shipment(self, overrides=None):
...
+        }
+        # Merge local overrides with passed-in overrides
+        if overrides:
+            overrides_local.update(overrides)
+        payload = self.fake_request("/mto-shipments", "post", PRIME_API_KEY, overrides=overrides_local)
+
+        headers = {"content-type": "application/json"}
+        resp = self.client.post(
+            prime_path("/mto-shipments"),
+            name=prime_path("/mto-shipments — expected failure"),
```
ronaktruss, line 407: Using an — (em dash) here because it better separates the url and makes it easier to search, but might not be the best approach as more people start adding to this repo. It might not be obvious that this is an em dash when reading the code and lead to  inconsistencies (hyphens being used in other places).

tasks/prime.py
gidjin: Should we add a `check_failed_response` that does similar logging but expects it to not be a 2XX code? Asking because it looks like `check_responses` is defined in `tasks/base.py`

tasks/prime.py
gidjin: It's too bad the `@tag`s don't show up in the report. I wonder if there is a way to do that.

tasks/prime.py
ronaktruss: Just added a change to modify `check_responses` to take an optional parameter of `expected_status_code` (defaults to 2xx) where it will do more verbose logging when the expected status code doesn't match the actual status code.

tasks/prime.py
```diff
@@ -324,6 +324,12 @@ def create_mto_service_item(self, overrides=None):
     @tag(MTO_SHIPMENT, "createMTOShipment")
     @task
     def create_mto_shipment(self, overrides=None):
+        def guarantee_unique_agent_type(agents):
+            agent_types = {agent["agentType"] for agent in agents}
```
duncan-truss, line 328: Nice, I realized you were doing a set operation below but missed at first that this coerces duplicates.

tasks/prime.py
```diff
@@ -362,6 +369,48 @@ def create_mto_shipment(self, overrides=None):
...
 
+    @tag(MTO_SHIPMENT, "createMTOShipment", "expectedFailure")
+    @task
+    def create_mto_shipment_with_duplicate_agents(self, overrides=None):
+        # If moveTaskOrderID was provided, get that specific one. Else get any stored one.
+        object_id = overrides.get("moveTaskOrderID") if overrides else None
+
+        move_task_order = self.get_stored(MOVE_TASK_ORDER, object_id)
+        if not move_task_order:
+            logger.debug("createMTOShipment: ⚠️ No move_task_order found")
```
duncan-truss, line 380: For consistency I think only errors are using the caution symbol.  It may make things confusing if debug messages also start using it.

ronaktruss: Right now `⚠️` is being used on both error logging and when we gracefully terminate the load test within the `prime.py` file. This is a case where we are gracefully terminating.

Are you suggesting that we change this throughout the file? Since we shouldn't expect data issues, I think they should be flagged when they occur. Having some sort of icon helps make this more transparent. While I think ⚠️ is an appropriate fit, I am not inherently opposed to having a different message and only using ⚠️ for the load testing error messages during a request.

duncan-truss: You're right I see there are some other debug statements that are using that icon now.  For me it's been helpful to only locate the errors in the logs.  I think most of these debug cases are early in the locust tests before there is any data to work on depending on the sequence of the tasks that are called and the users that are spawned.  I think the message is helpful but might detract from the real error responses.

duncan-truss (approved):
Create shipment looks fixed and failure endpoint looks good too!

If it's not a ton of trouble it looks like the `updateMTOAgent` might be failing for the same reason as the create shipment request was.  Did you see that as well?

ronaktruss:
@duncan-truss Yeah that does look like the case. I'll take a closer look on Monday

ronaktruss:
Just updated `updateMTOAgent`. I used a "naive" approach that just keep the type the same when doing an update. An alternative approach is to utilize the same agent type when there are two agents and allow for agent types to be different when there is only one agent. Planning on updating this to the alternative approach when I get some additional time to work on this later today.

There is also an ETag issue with this test. This requires some additional investigation and should likely separated out into its own story.

tasks/base.py
```diff
@@ -29,7 +30,10 @@ def check_response(response: Response, task_name="Task", request=None):
         logger.exception("Non-JSON response.")
         return None, False
 
-    if not str(response.status_code).startswith("2"):
+    if expected_status_code is None:
+        expected_status_code = "2"
```
duncan-truss, line 34: You could initialize the `expected_status_code` param to "2" instead of None, if you really wanted to avoid this check.

tasks/prime.py
```diff
@@ -350,6 +356,7 @@ def create_mto_shipment(self, overrides=None):
         if overrides:
             overrides_local.update(overrides)
         payload = self.fake_request("/mto-shipments", "post", PRIME_API_KEY, overrides=overrides_local)
+        guarantee_unique_agent_type(payload["agents"])
```
duncan-truss, line 359: I think I've seen it done elsewhere but it might not be obvious that this modifies the payload variable directly instead of returning a new dict.

tasks/prime.py
```diff
@@ -362,6 +369,48 @@ def create_mto_shipment(self, overrides=None):
...
+        object_id = overrides.get("moveTaskOrderID") if overrides else None
+
+        move_task_order = self.get_stored(MOVE_TASK_ORDER, object_id)
+        if not move_task_order:
+            logger.debug("createMTOShipment — expected failure: ⚠️ No move_task_order found")
+            return (
+                None  # we can't do anything else without a default value, and no pre-made MTOs satisfy our requirements
+            )
+
+        agent_type = random.sample(["RELEASING_AGENT", "RECEIVING_AGENT"], 1).pop()
```
duncan-truss, line 385: I think you can use random.choice here instead of sample with pop https://docs.python.org/3/library/random.html#random.choice

duncan-truss (approved):
🎉 Newest changes looks good!

PR #94, opened 2021-12-15
[MB-10651] Prevent create payment request from failing actual weight

NamibiaTorres:
## Description

The original ticket was to fix tests that were failing on create_payment_request due to missing actual_weights. When running load_testing, this particular failure was not found, however, there were issues with how data was being retrieved. To fix this issue, the code was updated so that the helpers could no longer call a part of the move, which is what they were doing before. This was leading to stale data, so instead the move id was retrieved and stored so that when a move is fetched, the original move is retrieved. 

While this may be inefficient, it was a workaround to get the payment_requests to pass again. Another ticket will be created to more efficiently store and retrieve data for load_testing.

Duncan's additions:
- Fix NTS weight being added to non-NTS shipments
- Add expected failure for payment requests for service items with missing shipment actual weights.

## Setup

1. First run the tests in the `master` branch.

Run the server in mymove:

```sh
make server_run
```
Run load_testing:

```sh
make load_test_prime
```

```sh
make load_test_prime_workflow
```
2. Repeat the steps above, but in this branch: `MB-10651-prevent-create-payment-request`

## Code Review Verification Steps

* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10651) for this change.

## Screenshots

Screen Shot 2021-12-27 at 2 55 56 PM

PR #95, opened 2021-12-15
[MB-10650] fix load test errors for update shipment status endpoint (support api)

pearl-truss:
## Description

The `update_mto_shipment_status` function in `tasks/prime.py` was consistently failing on the support endpoint for updating a shipment's status with 409 conflict error due to the fact that milmove restricts transitioning shipments between certain statuses. This PR updates the `update_mto_shipment_status` function so that it only attempts to update a shipments status is that incoming and transitioning statuses are in valid states

## Reviewer Notes

In the milmove repo, refer to [`pkg/services/mto_shipment/shipment_router.go`](https://github.com/transcom/mymove/blob/master/pkg/services/mto_shipment/shipment_router.go) for the code that determines which statuses a shipment can transition to.

## Setup
Follow set up instructions for running load testing locally (have server populated and running in the milmove repo) and then  in the load testing repo run `locust -f locustfiles/prime.py --host local`

Note: It can take awhile for this endpoint to get hit by the load test so you may want to comment out irrelevant test like (creating payment request, updating mto service items, etc)

Verify that the `PATCH | /support/v1/mto-shipments/{mtoShipmentID}/status` runs without conflict errors 
Screen Shot 2021-12-21 at 10 44 22 AM

* [ x] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10650) for this change.


locustfiles/prime.py
```diff
@@ -41,7 +41,7 @@ class SupportUser(MilMoveHostMixin, HttpUser):
     parser = support_api
 
     wait_time = between(0.25, 9)
-    tasks = {SupportTasks: 1}
+    tasks = {SupportTasks: 10}
```
pearl-truss, line 44: temp change, will remove when this is out of draft status

tasks/prime.py
```diff
@@ -230,36 +230,36 @@ def list_moves(self):
         # if this is running locally)
         self.set_default_mto_ids(moves or [])
 
-    @tag(MTO_SERVICE_ITEM, "createMTOServiceItem")
```
pearl-truss, line 233: Test commented out to better see when the update shipment status endpoint was being hit. They will be added back when I take this out of draft

tasks/prime.py
```diff
@@ -568,6 +587,8 @@ def update_mto_shipment_status(self, overrides=None):
             headers=headers,
             **self.user.cert_kwargs,
         )
+        if resp.status_code == 409:
+            print("QQQ ============================= \n" + "payload status: " + payload["status"] + "\n ======= shipment status:===" + mto_shipment["status"] + "mto shipment id: " + str(mto_shipment["id"]))
```
pearl-truss, line 591: One thing I'm seeing with the statement that I'm trying to debug is:
mto_shipment["status"] is coming back as SUBMITTED
but when I look at mto_shipment["id"] in the database, the status is APPROVED and the logic that starts on line 570 seems to be using the db status which causes an (expected) failure if the payload["status"] isn't DIVERSION_REQUESTED (the only valid status a mto_shipment can be updated to if it's current status is APPROVED)
Screen Shot 2021-12-15 at 5 24 03 PM
Screen Shot 2021-12-15 at 5 24 41 PM

ronaktruss: If I am understanding this correctly, it looks like this is an issue with the local storage making it so that `mto_shipment["status"]` is giving you stale data. If we make a change similar to the change in Namibia's [PR](https://github.com/transcom/milmove_load_testing/pull/94), we should be able to get the actual status and then add the appropriate logic to either coerce the data to pass or skip the test.

tasks/prime.py
```diff
@@ -615,24 +615,87 @@ def update_mto_shipment_status(self, overrides=None):
         if not mto_shipment:
             logger.debug("updateMTOShipmentStatus: ⚠️ No mto_shipment found.")
             return None  # can't run this task
+        # To avoid issues with the mto shipment being stale
+        # retrieve the move associated with the shipment
+        # and then use the newly fetched move to the find most up to date version of the shipment
+        move_id = mto_shipment["moveTaskOrderID"]
+        headers = {"content-type": "application/json"}
+        resp = self.client.get(
```
duncan-truss, line 623: Do you know the root cause of the failure? Is there another endpoint that isn't updating the local store properly?  Do we know it's not just a race condition depending on the number of users and moves?

tasks/prime.py
```diff
@@ -615,24 +615,87 @@ def update_mto_shipment_status(self, overrides=None):
...
+                )
+                new_mto_shipment, success = check_response(resp, "updateMTOShipmentStatus", payload)
+
+                if success:
+                    self.update_stored(MTO_SHIPMENT, mto_shipment, new_mto_shipment)
+                    return mto_shipment
+
+    @tag(MTO_SHIPMENT, "updateMTOShipmentStatus", "expectedFailure")
+    @task
+    def update_mto_shipment_with_invalid_status(self, overrides=None):
```
duncan-truss, line 672: It looks like this is getting run a lot more frequently than the real successful request (200 vs 40).  I wonder if we should try to limit the ratio of expected errors otherwise it may soon makeup a very large part of our requests.

tasks/prime.py
pearl-truss: I'm not 100% sure, I'm leaning more towards a race condition because when I checked if any test was wasn't updating the local store properly nothing surfaced. Even when I had most test commented out and it was still an issue

tasks/prime.py
pearl-truss: The number of times the successful request is run is artificially deflated since it first checks if the randomly generated shipment statuses are valid transitioning pairs and bails if they aren't. During the run that was captured in the screenshot, there was a range of 108-447

tasks/prime.py
duncan-truss: Yea in a previous PR I was commenting out other tests too and only running ~2 users to figure out if it was a bug or just random contention.  It turned out that the move eTag being returned was a bug in mymove in that case.  This will count in the stats to the support API GET move-task-order requests too.

tasks/prime.py
duncan-truss: It looks like we can provide a weight along with the @task decorator like `@task(1)`.  I don't know if it's valid to provide a non whole number like `.25` to maybe not run the expected failure task quite so much.

duncan-truss (approved):

tasks/prime.py
pearl-truss: Screen Shot 2021-12-27 at 12 15 22 PM

I couldn't use a float but I was able to adjust the other support task to make the expected failure task run relatively less frequently

PR #96, opened 2021-12-20
[MB-10649] Prime update shipment avoid geocode failures by using non-faker zipcode library

duncan-truss:
## Description

When running load testing a lot of errors in the logs were failures because of an address that was not able to be geocoded by the HERE maps API service.  Part of that issue is that the faker library that we use chooses a random 5 digit number between a start and end range of state ZIP codes.  This is inaccurate because ZIP codes are not contiguously valid and we ended up using ZIP codes that don't exist.

Milmove uses a static source of ZIP codes that originates for the GeoNames database for some form address validation.  Instead of porting that to be python formatted and write helper lookups, I found a library https://github.com/seanpianka/Zipcodes that seems comprehensive and would give us supplementary data such as the city.

I added some handling to the object fields so now address definitions should have a coherent city, state, and postal code value that match.

## Reviewer Notes

We may still see geocode failures until the HERE maps planner is removed/migrated to use ZIP distance instead of geocoding an address to latitude and longitude.

I don't think the data dump of the new library should impact memory usage but it may be something to keep an eye on.

If we wanted to lookup ZIPs to match one of the duty stations (for SC) or GBLOC (for TOO) we probably could but that might be a lot more domain logic to write than something that sticks to the API data types.  Those are also part of the move/orders and not something we have handy when performing shipment requests.  I left some comments on the Jira story that proposes that https://dp3.atlassian.net/browse/MB-10760

## Setup

From the mymove repo
```sh
make server_run
```

From the load testing repo:
```sh
make load_test_prime
```

To run tests I added:
```sh
pytest
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10649) for this change.

### DIFF
--- a/utils/constants.py
+++ b/utils/constants.py
@@ -51,6 +51,7 @@ class DataType(ValueEnum):
     POSTAL_CODE = "postalCode"
     POSTAL_CODE_VARIANT = "postal_code"
     COUNTRY = "country"
+    CITY_STATE_ZIP = "city_state_zip"  # a virtual field to find valid combinations of city, state, and zip
     DATE = "date"
     DATE_TIME = "date-time"  # inconsistent, but matches the swagger format name
     TIME_MILITARY = "timeMilitary"
--- a/utils/fake_data.py
+++ b/utils/fake_data.py
@@ -99,6 +99,17 @@ class MilMoveProvider(AddressProvider, DateProvider):
             if state != "AK" and state != "HI":
                 return random.choice(zipcodes.filter_by(state=state)).get("zip_code")
 
+    def safe_city_state_zip(self):
+        """
+        Returns a combination of a valid city, state, and postal code.
+        :return: tuple
+        """
+        while True:
+            state = self.state_abbr(include_territories=False)
+            if state != "AK" and state != "HI":
+                location = random.choice(zipcodes.filter_by(state=state))
+                return {"city": location["city"], "state": location["state"], "postalCode": location["zip_code"]}
+
     def safe_country(self):
         """
         Returns a safe country as a string. Currently restricting to US based CONUS locations not including territories
@@ -142,6 +153,7 @@ class MilMoveData:
             DataType.POSTAL_CODE: self.fake.safe_postal_code,
             DataType.POSTAL_CODE_VARIANT: self.fake.safe_postal_code,
             DataType.COUNTRY: self.fake.safe_country,
+            DataType.CITY_STATE_ZIP: self.fake.safe_city_state_zip,
             DataType.DATE: self.fake.date,
             DataType.DATE_TIME: self.fake.iso_date_time,
             DataType.TIME_MILITARY: self.fake.time_military,
--- a/utils/fake_data.py
+++ b/utils/fake_data.py
@@ -2,12 +2,14 @@
 """ utils/fake_data.py is for Faker classes and functions to set up test data """
 import logging
 import json
+import random
 import string
 from datetime import datetime
 
 from faker import Faker
 from faker.providers.date_time import Provider as DateProvider  # extends BaseProvider
 from faker.providers.address.en_US import Provider as AddressProvider  # extends BaseProvider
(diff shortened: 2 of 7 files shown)


utils/constants.py
```diff
@@ -51,6 +51,7 @@ class DataType(ValueEnum):
     POSTAL_CODE = "postalCode"
     POSTAL_CODE_VARIANT = "postal_code"
```
duncan-truss, line 52: I don't think this variant is ever part of an address object anymore so I was able to ignore it I think.

utils/fake_data.py
```diff
@@ -89,12 +91,31 @@ def safe_street_address(self):
...
+            if state != "AK" and state != "HI":
+                location = random.choice(zipcodes.filter_by(state=state))
+                return {"city": location["city"], "state": location["state"], "postalCode": location["zip_code"]}
+
+    def safe_country(self):
+        """
+        Returns a safe country as a string. Currently restricting to US based CONUS locations not including territories
+        in the states list.
+        """
+        return "United States of America"
```
duncan-truss, line 118: I was seeing other countries being used with US address fields so I'm restricting this to the US with CONUS only moves.

utils/fields.py
```diff
@@ -262,10 +262,23 @@ def generate_fake_data(self, faker, overrides=None, require_all=False):
         if self.discriminator and d_value:
             overrides[self.discriminator] = d_value
 
+        # Address fields are related so we want to have city, state, zip values that are valid instead of independent
+        # random ones.  This relies on the swagger definition using the name "Address" and the properties being named
+        # the same across APIs.
+        virtual_city_state_zip = None
+        if self.is_address_object():
```
duncan-truss, line 269: Unfortunately once the API is parsed we have no idea if say a field like `pickupAddress` referenced the `Address` definition in the `$ref` property.  I am testing for the required fields instead so that if a new address field gets added we don't need an allowlist of all possible property names like `destinationAddress`, ...

utils/fields.py
```diff
@@ -262,10 +262,23 @@ def generate_fake_data(self, faker, overrides=None, require_all=False):
...
+            virtual_city_state_zip = BaseAPIField(
```
duncan-truss, line 270: My solution here was to not subclass the object field just to handle the address definition because it would have been a lot of redundancy and boilerplate to copy and it didn't seem to fit with how we are creating our class types.

I created a virtual object field `city_state_zip` that will return the 3 valid values at once instead of how each used to get called individually and not match one another.

utils/tests/test_fields.py
```diff
@@ -537,3 +537,80 @@ def test_generate_fake_data_discriminator(self, mocker):
...
+        fake_data = self.object_field.generate_fake_data(self.faker)
+
+        city_state_zip = self.MOCK_FAKE_DATA[DataType.CITY_STATE_ZIP]
+
+        # require_all=False, so we should only see required fields:
+        assert fake_data == {
+            "streetAddress1": self.MOCK_FAKE_DATA[DataType.STREET_ADDRESS],
+            "city": city_state_zip["city"],
+            "state": city_state_zip["state"],
+            "postalCode": city_state_zip["postalCode"],
```
duncan-truss, line 615: This asserts that we get back the city_state_zip fake data instead of the individual field fake data.

utils/constants.py
pearl-truss: I think it still exists in the internal api

duncan-truss: Yep it's still in some places in the internal API but not part of any Address definition more one off fields.  It should use the new library too for those individual fields https://github.com/transcom/milmove_load_testing/pull/96/files#diff-13ff359e13fc9a6114896673b1c3cbb9ad704f9e7a5a3818f0cbe5786cfc2848R100

pearl-truss (approved):
There's a failing circleci issue (may not be related to these changes) but otherwise LGTM

PR #97, opened 2021-12-22
MB-10533 Locust best practices refactors

felipe-lee:
## Description

Implementing some of my suggestions for best practices refactors. Key points are:
* Use a `RestTaskSet` class that can handle automatically parsing json and will auto-fail if the response isn't json. 
    * This also gives us the ability to mark a task as a success or failure, regardless of the status code.
    * Note that this is different from the original plan to use `RestHttpUser` because we don't really need the things it provides at a user level, we need them at a task set level since we define all our load tests in task sets.
* Make it easier to make requests with all the necessary bits (proper url, certs, etc.). To do this, I've created some helper functions that give you what you need, which can then be used with the `client`, or with our own thing, e.g. we can use the `requests` library directly.
* Create some parent classes we can easily use without having to worry about the proper order for mixins to be in.
    * Only have one so far, but I'm going to take a look to see if anything else would be good to do. 
* Pull cert creation/deletion out of the user classes and instead do it at the locust init/shutdown level. This simplifies it a bit. Also cut down on the ability to run against arbitrary environments since we are only using local and dp3.

## Setup

Things are still in progress, but I added a locustfile called `rest.py` that shows a very simple usage of the new `RestHttpUser` class. You can run this file by starting up the mymove server, and then running this command:

```sh
locust -f locustfiles/rest.py --host local
```

If you'd like to run without the UI, here's the command with some extra args:

```sh
locust -f locustfiles/rest.py --headless --host local -u 100 -r 5 -t 30s -T listMoves
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10533) for this change.
* [Locust Best Practices Wiki](https://dp3.atlassian.net/wiki/spaces/MT/pages/1582891038/Locust+Best+Practices)
    * I need to update this wiki though with some of my learnings as I've tried to implement changes...


locustfiles/lifecycle.py
```diff
@@ -0,0 +1,106 @@
+# -*- coding: utf-8 -*-
+"""
+This example shows the various ways to run things before/outside the normal task execution
+flow, which is very useful for fetching test data.
+
+This was copied from https://github.com/locustio/locust/blob/master/examples/test_data_management.py
```
duncan-truss, line 6: It looks like both our repo and theirs are MIT licensed, maybe we should include their copyright as well in this file https://github.com/locustio/locust/blob/master/LICENSE

felipe-lee: Tbh, I don't know much about the license stuff. Do you mean like include a link to their license? Or add a copyright statement?

duncan-truss: I would say add the copyright statement to the top of that file if that's what you copied here.

utils/hosts.py
```diff
@@ -62,7 +53,7 @@ def host_name(self, env, is_api=False, port="0000", protocol="https"):
         # allow us to point to another domain if we need to
         base_domain = os.getenv("BASE_DOMAIN", "loadtest.dp3.us")
         # NOTE: deployed protocol is always https
-        return f"https://{'api' if is_api else self.deployed_value}.{base_domain}"
+        return f"https://api.{base_domain}"
```
ronaktruss, line 56: In the previous iteration, `is_api` was only true for the prime tests. So when running the office tests we wouldn't expect the `api` prefix as part of the host name. Is this still the case/ is there a related change that modifies the way this function is used.

felipe-lee: 😞 I swear I had seen it on the other ones as well, but now that you called it out and I'm looking on master, you're right. I'll bring this back in. That's what I get for primarily only testing the prime api with my changes. Thanks for catching this!

felipe-lee: I kind of reverted it. I didn't do a direct revert because `is_api` is somewhat confusing since they're all APIs. 

Here is the commit that brings this back: https://github.com/transcom/milmove_load_testing/pull/97/commits/3bd1f6624afa560c4b619eff358f4f62549bfb73

Though I am also thinking of maybe changing how URLs are formed to be this way: https://github.com/transcom/milmove_load_testing/pull/97/commits/8196f38fa91ce6f8a43af316b3cba2400a12cb22

tasks/prime.py
```diff
@@ -835,7 +920,13 @@ def update_mto_service_item_status(self, overrides=None):
             logger.debug("updateMTOServiceItemStatus: ⚠️ No mto_service_item found")
             return None
 
-        payload = self.fake_request("/mto-service-items/{mtoServiceItemID}/status", "patch", SUPPORT_API_KEY, overrides)
+        payload = fake_data_generator.generate_fake_request_data(
+            api_key=APIKey.SUPPORT,
+            path="/mto-service-items/{mtoServiceItemID}/status<",
```
ronaktruss, line 925: Extra `<` character in the path

ronaktruss (commented):
LGTM. I think there are areas that are a bit confusing at the moment since we are supporting both the old way and the new way. Specifically the way that urls are setup in the MilMoveHostMixin, which to my understanding isn't used in the new format you are proposing.

tasks/prime.py
felipe-lee: Oh good catch, thanks!

felipe-lee:
> LGTM. I think there are areas that are a bit confusing at the moment since we are supporting both the old way and the new way. Specifically the way that urls are setup in the MilMoveHostMixin, which to my understanding isn't used in the new format you are proposing.

Yeah, I'm going to see how much work it would be to just switch things over so we don't have to live in both worlds. Based on my earlier testing, it should hopefully not be too much 🤞🏼

felipe-lee:
I've written tests for all the helpers and such I wrote. At least I think I have... 🤔 ...but it's also the end of the day so I may have missed something. At this point, the main thing left is to address what @ronaktruss pointed out about it being confusing at the moment because we have both the new and old ways of doing things. I'll tackle that on monday, and depending on how much it is, might open a second PR to stop this one from getting too much bigger. We'll see. For now though, the core things that I've added I think are ready for review.

utils/auth.py
```diff
@@ -0,0 +1,54 @@
...
+        deployed_tls_cert = os.environ[f"MOVE_MIL_{host_upper}_TLS_CERT"]
+        deployed_tls_key = os.environ[f"MOVE_MIL_{host_upper}_TLS_KEY"]
+    except KeyError:
+        logger.debug(f"Unable to find cert and key values for environment: {MilMoveEnv}")
+
+        raise ImplementationError(
+            "Cannot run load testing in a deployed environment without the matching certificate and key."
+        ) from None
+
+    with open(DP3_CERT_KEY_PEM, "w") as f:
```
duncan-truss, line 35: Does hardcoding this prevent us from running against say experimental?

felipe-lee: To be honest, I'm not sure. In the past I would have said yes, but since experimental got moved to `dp3`, it might be that our cert key might work with either one. But it would require us to edit other code like the bit about which base domain to point to.

I wasn't too concerned since my understanding is that we really only want to run against the load testing environment. We only used experimental before because we didn't have a dedicated environment. Do you think we should keep the ability to point at other environments though?

felipe-lee:
Note that I'm still working through updating the README, but as I was doing code examples for it, I realized we really don't need the `RestHttpUser` class, but rather that functionality at a `TaskSet` level since we define all our load tests in task sets.

utils/auth.py
duncan-truss: Oh I had forgotten that exp moved outside of the ATO boundary, I was mainly thinking we'd want to know if load testing performs differently inside the ATO infra.

It's a bit weird that we expose the host field in locust as configurable/overideable but make slightly different assumptions here.

No action needed.

felipe-lee: Yeah that is one weird change this introduces in that we don't really use the locust host for much other than as a starting point. Normally the locust host is used in the background by the locust client to create the full URL, but these changes make it so that the host is just a base string that we start off with and then we pass the client full urls that our `MilMoveRequestPreparer` sets up

README.md
```diff
@@ -1,194 +1,239 @@
...
 
 ## Overview
 
-MilMove is a system to help service members (and other authorized personnel) move their gear and possessions from one
-place to another. This codebase has been written to perform load tests on the MilMove app for the purpose of gathering
+MilMove is a system to help service members (and other authorized personnel) move their gear and
+possessions from one place to another.
+
+This codebase has been written to perform load tests on the MilMove app for the purpose of gathering
 data about responses times, finding breakpoints, and assessing the overall health of the system.
```
duncan-truss, line 76: It might be nice to say that we are only really testing the APIs and backend of the system.  It's not giving us insight into anything browser or client side.

README.md
```diff
@@ -1,194 +1,239 @@
...
+provides a number of [config options](https://docs.locust.io/en/stable/configuration.html) to allow
+you to manipulate which users and/or tasks run from any given locustfile.
 
-### `tasks/`
+In particular, our `locustfiles/all.py` combines all our `User` classes into a single run.
 
-Each `User` class needs a set of tasks to complete to be able to run a load test. All tasks are callables that can be
-manually set into the `tasks` attribute of the user, or they can be organized into instances of the locust class
-`TaskSet` and then associated with a user. This directory contains all of the tasks used in our load tests.
+### `scripts`
```
duncan-truss, line 103: just for consistency this is the only folder without a trailing slash `scripts/`

README.md
felipe-lee: Ah good point, I'll add that.

README.md
felipe-lee: oh good catch, thanks!

README.md
```diff
@@ -1,194 +1,239 @@
...
+3. Now run
 
-This should install the dependencies via `pipenv` automatically.
+    ```shell
+    direnv allow
+    ```
 
-If the base dependencies change, you can always re-run this command.
+   1. This should install the dependencies via `pipenv` automatically.
+   2. If the base dependencies change, you can always re-run this command.
```
duncan-truss, line 168: I think this is true the first time but I was generally having to run `pipenv install` unless I deactivated or destroyed the virtual environment.

felipe-lee: Ah interesting, yeah it looks like direnv only runs the install if there isn't already a venv: 
https://github.com/direnv/direnv/blob/f68f741895754e9551059c8e0132a796ad1c3c3a/stdlib.sh#L882-L904

I've updated the README with new instructions for updating python dependencies. Thanks for pointing this out!

locustfiles/lifecycle.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+@events.test_stop.add_listener
+def _(environment: Environment, **_kwargs):
+    print("7. stopping test run")
+    print(f"\n{environment.__dict__=}\n{_kwargs=}\n")
+
+
+class MyUser(HttpUser):
+    host = "https://postman-echo.com"
```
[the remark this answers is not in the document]
felipe-lee: Me either, but it's pretty cool for testing things.

utils/tests/test_users.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+"""
+Tests for utils/users.py
```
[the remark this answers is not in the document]
felipe-lee: I think this should be gone after this commit: https://github.com/transcom/milmove_load_testing/pull/97/commits/1c05f0d53cbdc31f7deabb360d681818394ac48d

README.md
```diff
@@ -225,85 +286,90 @@ To run a specific test, use:
...
+### Load Tests
 
-For more instructions and examples, please read [pytest's documentation](https://docs.pytest.org/en/stable/).
+There are several ways you can run load tests. You can run `locust` locally or use the deployed
+version. You can also use a local mymove server or the one deployed in the load test environment.
+This means you have three possible combinations:
 
-## Running Load Tests
+* Run `locust` locally against local mymove server
+* Run `locust` locally against load test env mymove server
```
duncan-truss, line 308: I might spell out loadtest.dp3.us here, I think it gets a little muddled with all the environment, host, and server terms being thrown around getting used.

felipe-lee: Does https://github.com/transcom/milmove_load_testing/pull/97/commits/1c32c1dd7eca887b8691dc4dce43b01ebf66f8fa help? Or do you mean some other way?

duncan-truss: Thanks that's exactly what I meant.

felipe-lee:
Ok, the docs are done I think, at least based on where the code is now. I'll work on seeing what it takes to switch existing load tests over to use the new stuff in a separate PR to keep it more manageable.

.editorconfig
```diff
@@ -9,6 +9,7 @@ end_of_line = lf
 indent_size = 2
 indent_style = space
 insert_final_newline = true
+max_line_length = 100
```
ronaktruss, line 12: Is there any additional setup that we need to do to get this to work? All of the other rules seem to be working fine on my end other than this rule.

felipe-lee: I don't think so, though maybe I should move it to just be for Markdown. I would like this to be set for python but didn't set it in `black`'s settings because it would have led to a lot of noise in this pr. Is it not working for any files for you? Or do you mean it isn't working for python files?

ronaktruss: Yeah I saw the details of your commit msg indicating the reason for not including in the `black` settings. For me its not working in the markdown files for some reason. (I was trying to do a local diff since the current one on github is hard to parse and wasn't able to get it to work.)

felipe-lee: Weird, I wonder why. You use VS code right?

felipe-lee: Oh I see, apparently it isn't implemented in the VS code plugin because they don't think it should be part of it: https://github.com/editorconfig/editorconfig-vscode/issues/53

felipe-lee: Hmm I guess maybe I should remove it completely then, otherwise we'll keep having differences between people using vscode and people using jetbrains IDEs

ronaktruss: Yeah, I'm using VSCode. What's the rationale behind having max line lengths for md files? I've generally been a fan of using wrapping as it makes the diffs cleaner imo.

ronaktruss (approved):
LGTM. I reviewed it with a slightly wider tooth comb than usual (given the length and having looked at this during the drafts), so I'd recommend waiting for a second approval if possible.

.editorconfig
felipe-lee: Mainly being able to have multiple things up side by side. I had the plain README, the preview, and a code window all side by side to help as I was writing and with the original super long lines it was harder to edit. Wrapping sometimes works weirdly, but I can change to using other screens or something.

utils/hosts.py
```diff
@@ -2,57 +2,54 @@
...
         """
         if isinstance(env, MilMoveEnv):
             env = env.value  # ensure that we're using the value string instead of the Enum literal
 
-        if env not in MilMoveEnv.values():
-            raise ImplementationError("The environment for determining the host name must be included in MilMoveEnv.")
+        milmove_env = MilMoveEnv(value=env)
 
-        if env == MilMoveEnv.LOCAL.value:
+        if milmove_env == MilMoveEnv.LOCAL:
```
duncan-truss, line 52: I think you have a `milmove_env.is_local()` method you can call here?

utils/request.py
```diff
@@ -0,0 +1,236 @@
...
+    # The environment the host is running in (ex. locally or deployed).
+    # Can be any of the values in MilMoveEnv.
+    env: MilMoveEnv
+
+    request_preparer: MilMoveRequestPreparer
+
+    def __init__(self: Union[TaskSet, "MilMoveRequestMixin"], *args, **kwargs) -> None:
+        """
+        Sets up the env based on value from the --host flag. Note that the `self.host` value is set
+        on the class from the command line flag BEFORE initialization.
```
duncan-truss, line 219: This is right but could be overridden through the host form field if they start a new load test through the UI.

felipe-lee: Ah good point, I'll update this.

utils/hosts.py
felipe-lee: Ah, thanks for catching that

duncan-truss (approved):
New patterns look good to me, thanks for taking this on.

I ran into some issues on the dp3 load test environment but we can look into those after this is merged.

PR #98, opened 2021-12-29
[MB-10562] Create dependabot config for pipenv and docker

duncan-truss:
## Description

This PR creates a dependabot config in `.github/dependabot.yml` to run daily at 11:00 (the same as mymove) for both the pipenv dependencies and the docker dependency.

I will need to update the [dependencies documentation](https://dp3.atlassian.net/wiki/spaces/MT/pages/1419870254/Dependency+Updates+Process+Working+Document) to make sure these are taken care of with the others.  Since GitHub acquired Dependabot I noticed some [broken links in the mymove-docs repo](https://transcom.github.io/mymove-docs/docs/tools/dependabot/manage-dependabot) too redirected to the wrong places.

Once merged we can manage dependabot from the insights -> dependency graph -> dependabot page https://github.com/transcom/milmove_load_testing/network/updates

[Dependabot configuration options](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates)

[mymove Dependabot reference file](https://github.com/transcom/mymove/blob/master/.github/dependabot.yml)

## Reviewer Notes

I don't think this is a PR that can be tested until it's merged.

We may decide to add base versions in our pipfile instead of wildcards for some dependencies.

I can't remember how we settled on the current python version but I could see us turning off the docker checks so it doesn't get out of line with pipenv.

## Setup

```sh
echo "Code goes here"
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10562) for this change.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,20 @@
+version: 2
+updates:
+  # pipenv requires using the "pip" value https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#package-ecosystem
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: daily
+      time: "11:00"
+    open-pull-requests-limit: 10
+    labels:
+      - dependencies
+  # The python version base image is the only real docker dependency but should match what's in the pipfile
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: daily
+      time: "11:00"
+    open-pull-requests-limit: 10
+    labels:
+      - dependencies


felipe-lee:
> I don't think this is a PR that can be tested until it's merged.

Yeah I think you're right, though I also don't have much experience with dependabot.

> We may decide to add base versions in our pipfile instead of wildcards for some dependencies.

Yeah, this might be good because otherwise we're only really going to have the lock file to go by and that'd probably be annoying to read through when seeing the version updates. Do you think we should do that in this PR? Or wait and see if it really is an issue?

> I can't remember how we settled on the current python version but I could see us turning off the docker checks so it doesn't get out of line with pipenv.

We didn't really have a set way. Different people would just test newer python versions if they felt like it and if they seemed to work, committed the change to the version. I feel like this could serve to point out when we should maybe check to see if a new python version will work.

felipe-lee:
> I will need to update the [dependencies documentation](https://dp3.atlassian.net/wiki/spaces/MT/pages/1419870254/Dependency+Updates+Process+Working+Document) to make sure these are taken care of with the others. Since GitHub acquired Dependabot I noticed some [broken links in the mymove-docs repo](https://transcom.github.io/mymove-docs/docs/tools/dependabot/manage-dependabot) too redirected to the wrong places.

Related to my last comment above, I wonder if as you update this documentation, there should be a mention of how and when to check for python version updates.

felipe-lee (approved):
LGTM, at least based on how it seems. Like we mentioned in the description/comments, we probably need to merge to test it and this seems merge-able to me

duncan-truss:
> > I will need to update the [dependencies documentation](https://dp3.atlassian.net/wiki/spaces/MT/pages/1419870254/Dependency+Updates+Process+Working+Document) to make sure these are taken care of with the others. Since GitHub acquired Dependabot I noticed some [broken links in the mymove-docs repo](https://transcom.github.io/mymove-docs/docs/tools/dependabot/manage-dependabot) too redirected to the wrong places.
> 
> Related to my last comment above, I wonder if as you update this documentation, there should be a mention of how and when to check for python version updates.

@felipe-lee Yea the docker notification may be a decent proxy for that, I know a drawback of Dependabot noted elsewhere was that it doesn't do major version upgrades.

PR #99, opened 2021-12-29
Mb 10647 rp create moves

ronaktruss:
## Description

Adds move setup as part of the prime load tests.

## Reviewer Notes

The code is a giant block of code at the moment and likely will be worthwhile to separate out where the comments are made for each distinct step. I have tentatively refrained from doing this because there are some potential changes with how/where this will be used in [Felipe's PR](https://github.com/transcom/milmove_load_testing/pull/97) and the unknowns of how and if these functions will be used by other tests. In addition to those reason, we are also utilizing some of the data from previous steps so this will require some level of data management / parameters to refactor.

## Setup

When running the prime load tests using
```sh
make load_test_prime
```
you should be able to log into the the most recently created user to see that they have successfully created and submitted a move. (If you don't have a clean DB you can optinally add `print(email)` to the end of the  `on_start` function to see the email and login using that email to verify.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10647) for this change.

### DIFF
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -260,7 +260,6 @@ class PrimeTasks(PrimeDataStorageMixin, ParserTaskMixin, CertTaskMixin, TaskSet)
             name="/internal/service_members/{serviceMemberId}",
             data=json.dumps(payload),
             headers={"content-type": "application/json"},
-            **self.user.cert_kwargs,
         )
 
         overrides = {"permission": "NONE"}
@@ -272,7 +271,6 @@ class PrimeTasks(PrimeDataStorageMixin, ParserTaskMixin, CertTaskMixin, TaskSet)
             name="/internal/service_members/{serviceMemberId}/backup_contacts",
             data=json.dumps(payload),
             headers={"content-type": "application/json"},
-            **self.user.cert_kwargs,
         )
 
     @tag(MOVE_TASK_ORDER, "listMoves")
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -293,7 +293,7 @@ class PrimeTasks(PrimeDataStorageMixin, ParserTaskMixin, CertTaskMixin, TaskSet)
         payload = self.fake_request("/orders", "post", INTERNAL_API_KEY, overrides, True)
         order_resp = self.client.post(
             self.customer_path("/internal/orders"),
-            data=json.dumps(overrides),
+            data=json.dumps(payload),
             headers={"content-type": "application/json"},
         )
         order = order_resp.json()
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -257,9 +257,8 @@ class PrimeTasks(PrimeDataStorageMixin, ParserTaskMixin, CertTaskMixin, TaskSet)
             "email_is_preferred": True,
             "current_station_id": current_station_id,
         }
-
         payload = self.fake_request("/service_members/{serviceMemberId}", "patch", INTERNAL_API_KEY, overrides, True)
-        self.client.patch(
+        service_member_resp = self.client.patch(
             self.customer_path(f"/internal/service_members/{service_member_id}"),
             name="/internal/service_members/{serviceMemberId}",
             data=json.dumps(payload),
@@ -306,6 +305,31 @@ class PrimeTasks(PrimeDataStorageMixin, ParserTaskMixin, CertTaskMixin, TaskSet)
             self.customer_path(f"/internal/uploads?documentId={document_id}"),
             name="/internal/uploads",
             files=upload_file,
+        )
+
+        # Setup customer shipment
+        move_id = order["moves"][0]["id"]
+        self.client.patch(
+            self.customer_path(f"/internal/moves/{move_id}"),
+            name="/internal/moves/{moveId}",
+            data=json.dumps({"selected_move_type": "HHG"}),
+            headers={"content-type": "application/json"},
+        )
+
+        service_member = service_member_resp.json()
+        address = service_member["residential_address"]
+        address.pop("id")  # remove unneeded id
+        overrides = {
+            "moveTaskOrderID": move_id,
+            "shipmentType": "HHG",
+            "pickupAddress": address,
+            "agents": [],
+        }
+        payload = self.fake_request("/mto_shipments", "post", INTERNAL_API_KEY, overrides, True)
+        self.client.post(
+            self.customer_path("/internal/mto_shipments"),
+            data=json.dumps(payload),
+            headers={"content-type": "application/json"},
         )
 
     @tag(MOVE_TASK_ORDER, "listMoves")
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -133,7 +133,7 @@ class PrimeDataStorageMixin:
 
         data_list.append(new_data)
 
-    def set_default_mto_ids(self, moves):
+    def set_default_mto_ids(self, move_id):
         """
         Given a list of Move Task Orders, gets the four ID values needed to create more MTOs:
           - contractorID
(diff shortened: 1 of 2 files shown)


felipe-lee (commented):
Nice, thanks for putting this together, it'll hopefully get us closer to making it easier to run load tests.

One question I have is that is the expectation then that when the load tests run, they'll search for all moves and be able to grab this one? I mainly want to check if the plan was that or for this move ID to be shared like the other things are shared atm. I know we've talked about changing how we handle shared stored data, so I think the way it is now is good, but want to check what your intention was.

I agree that it'd be good to break this out and agree that my changes would affect how it's done so maybe good to do that in a follow-up PR.

I would like there to be error handling in case any of the requests fail, but I'm fine with that being part of the follow-up PR that puts these in helper functions.

ronaktruss:
@felipe-lee My understanding of the existing code is that we are just checking existing moves to get a move which we can extract the `contractorID`, `uploadedOrdersID`, `destinationDutyStationID`, and `originDutyStationID`. With the way the code is written right now we are not guaranteeing that the newly created move is used to populate these values. 

The added code is only useful in the instance that we don't have any existing moves in a non local environment(for local we have some overrides) or if we want to populate additional customers to the DB. Part of the reason that we wanted to add this code was to ensure that the tests would work when run on a fresh DB. If we don't necessarily care about the creation of additional customers as part of this tests, we likely want to add a check to see if its necessary and to move this code/setup of the `default_mto_ids` to only occur once per test.

felipe-lee:
> @felipe-lee My understanding of the existing code is that we are just checking existing moves to get a move which we can extract the `contractorID`, `uploadedOrdersID`, `destinationDutyStationID`, and `originDutyStationID`. With the way the code is written right now we are not guaranteeing that the newly created move is used to populate these values.
> 
> The added code is only useful in the instance that we don't have any existing moves in a non local environment(for local we have some overrides) or if we want to populate additional customers to the DB. Part of the reason that we wanted to add this code was to ensure that the tests would work when run on a fresh DB. If we don't necessarily care about the creation of additional customers as part of this tests, we likely want to add a check to see if its necessary and to move this code/setup of the `default_mto_ids` to only occur once per test.

Awesome, no it sounds good as is right now. We can adjust in the future if needed.

felipe-lee (approved):

tasks/prime.py
```diff
@@ -242,8 +244,10 @@ def on_start(self):
         email = json_resp["email"]
         user_id = json_resp["id"]
 
-        origin_duty_stations = self.client.get(self.customer_path("/internal/duty_stations?search=29"))
-        current_station_id = origin_duty_stations.json()[0]["id"]
+        # Setup customer profile
+        duty_stations = self.client.get(self.customer_path("/internal/duty_stations?search=palms"))
```
duncan-truss, line 248: Thanks for making this more explicit.  I do know the default GBLOC of the office users being created is now `KKFA`.  I think the origin duty station matters for the services counselor but not for the TOO.  I don't think Twenty Nine Palms provides services counseling so this move should proceed to the TOO queue.

We will want the shipment's postal code to map to the KKFA gbloc probably though so it appears in that TOO user's queue.

tasks/prime.py
```diff
@@ -272,7 +274,80 @@ def on_start(self):
...
+        # Setup customer order
+        overrides = {
+            "service_member_id": service_member_id,
+            "issue_date": datetime.now().strftime("%Y-%m-%d"),
+            "report_by_date": datetime.now().strftime("%Y-%m-%d"),
+            "orders_type": "PERMANENT_CHANGE_OF_STATION",
+            "orders_type_detail": "HHG_PERMITTED",
+            "has_dependents": False,
+            "spouse_has_pro_gear": False,
+            "new_duty_station_id": stations[1]["id"],
```
duncan-truss, line 288: Do we know what this duty station will be as the second result?

tasks/prime.py
```diff
@@ -272,7 +274,80 @@ def on_start(self):
...
-            **self.user.cert_kwargs,
+        )
+
+        # Setup customer order
+        overrides = {
+            "service_member_id": service_member_id,
+            "issue_date": datetime.now().strftime("%Y-%m-%d"),
+            "report_by_date": datetime.now().strftime("%Y-%m-%d"),
+            "orders_type": "PERMANENT_CHANGE_OF_STATION",
+            "orders_type_detail": "HHG_PERMITTED",
```
duncan-truss, line 285: I don't think this field gets a value during customer onboarding, only when the office user updates the orders later.

tasks/prime.py
```diff
@@ -272,7 +274,80 @@ def on_start(self):
...
+            "has_dependents": False,
+            "spouse_has_pro_gear": False,
+            "new_duty_station_id": stations[1]["id"],
+            "orders_number": None,
+            "tac": None,
+            "sac": None,
+            "department_indicator": None,
+        }
+        payload = self.fake_request("/orders", "post", INTERNAL_API_KEY, overrides, True)
```
duncan-truss, line 294: To Felipe's point I would say at least after every request if we could check that it didn't error before using the result that would be helpful to know.

tasks/prime.py
```diff
@@ -272,7 +274,80 @@ def on_start(self):
...
+            headers={"content-type": "application/json"},
+        )
+
+        service_member = service_member_resp.json()
+        address = service_member["residential_address"]
+        address.pop("id")  # remove unneeded id
+        overrides = {
+            "moveTaskOrderID": move_id,
+            "shipmentType": "HHG",
+            "pickupAddress": address,
```
duncan-truss, line 325: This postal code will determine which TOO queue it gets routed to under the recent changes.

tasks/prime.py
```diff
@@ -242,8 +244,10 @@ def on_start(self):
...
         user_id = json_resp["id"]
 
-        origin_duty_stations = self.client.get(self.customer_path("/internal/duty_stations?search=29"))
-        current_station_id = origin_duty_stations.json()[0]["id"]
+        # Setup customer profile
+        duty_stations = self.client.get(self.customer_path("/internal/duty_stations?search=palms"))
+        stations = duty_stations.json()
+        current_station_id = stations[0]["id"]
 
         overrides = {
```
duncan-truss, line 252: I wonder if it's worth adding a custom personal_email or first_name/last_name values so that we can easily locate these load test users in the database.  I know it won't impact what shows up on the login screen since that is set during user creation and is only editable through the admin.

tasks/prime.py
ronaktruss: The first duty station is `MCAGCC Twentynine Palms` and the second is `29 Palms CA`. When running the service locally you can go to `http://milmovelocal:3000/internal/duty_stations?search=palms`

tasks/prime.py
ronaktruss: Nice catch. Yeah it doesn't look like the customer flow sends this

tasks/prime.py
ronaktruss: I think this is worth thinking about. Is there a specific issue or use case that we are trying to solve? When running locally I don't this matters all that much as their are other alternatives that the user can decide which method works for them (eg. printing the email, providing an email in the overrides). When doing this in a deployed environment, I think it depends on why someone may want this information. If multiple users are created in a single test specifying a single email may not actually help narrow down the selection. 

I'd lean towards not adding this functionality unless there is a specific need/issue that we are currently trying to solve. If there is a common issue we can default to a specific methodology.  I think this functionality would make sense a future state when this functionality is explicitly needed.

ronaktruss: Ie. When would one want to look at the customer from the DB? The load tests that we are running don't necessarily modify the move that we create in this setup flow.

duncan-truss: That's true it won't be very useful if this becomes the basis for customer load tests because there will be thousands or so.  Right now it's just a single move in the system so yea I think it might be nice to have a stable identifier of some sort. I also wanted to have a way to distinguish potential data already in the system from load test created moves.  When a request fails it does log the payload when running locally so you can work backwards from the ids.

ronaktruss:
@duncan-truss thanks for calling this out! I'll add a call to make the newly added move available to the prime. I was initially under the impression that we were fetching all moves using the support api, but we are in fact using the prime api to get the moves list.

ronaktruss:
Added some changes to refactor how we are setting the local cache for default mto ids.

Note: the search for duty stations should still work with the removal of `MCAGCC Twentynine Palms` ([ref](https://github.com/transcom/mymove/pull/7977)) as the search query still provides a few responses and we are just defaulting to the first and second choices though this may impact gbloc (not a concern for prime tests but may be for the office flow).

tasks/tests/test_prime.py
```diff
@@ -257,14 +260,16 @@ class PrimeSessionStorage2(PrimeDataStorageMixin):
         session_storage1 = PrimeSessionStorage1()
         session_storage2 = PrimeSessionStorage2()
 
-        session_storage1.set_default_mto_ids(test_moves)
+        for move in test_moves:
+            session_storage1.set_default_mto_ids(move["id"])
 
         assert session_storage1.default_mto_ids == expected_ids
         assert session_storage1.default_mto_ids == session_storage2.default_mto_ids
         assert len(responses.calls) == 4
```
duncan-truss, line 268: My understanding here, looking at the test moves, is that none have all of the mto_ids, which is why it still makes 4 requests?

I guess this isn't true in practice now with our single move but it probably still tests the method more thoroughly.

duncan-truss (approved):
LGTM!

Do we want every Prime user to create a move with `on_start` once the first one successfully saves the needed MTO IDs? Or do you think that's desirable once we can make them available to the prime instead of the current support tasks creating moves?

tasks/tests/test_prime.py
ronaktruss: Yeah with the change I made we likely won't and shouldn't need all 4 requests since our setup should make it so that all of the info is available in the first move. I kept this there as this behavior is still supported in the case that we run the on start function multiple times and for w/e reason there are some moves that don't properly create moves.

ronaktruss:
@duncan-truss We likely don't want to call the on_start method as is for multiple users (I didn't include the change in this PR because the changes in Felipe's PR will influence how we do that) There are two approaches that I was considering: adding a check to see if the default values area already set before doing the customer logic to create the move or leveraging the existing support apis to create moves and populate the other local variables so that we are not reliant on having multiple users for the tests to run properly. (Right now if we only have 1 user only the prime User is run and most of the tests don't actually run because they are missing move info and the tests are skipped)

PR #108, opened 2022-01-06
Rp update duty location

ronaktruss:
## Description

Changing the default origin duty location for moves created within the prime setup to belong in the `KKFA` GBLOC

## Reviewer Notes

## Setup

Reset local DB (in mymove repo):
```
make db_dev_reset
make db_dev_migrate
```

Run local back-end server (in mymove repo):
```
make client_run
```

Run front-end server (in mymove repo):
```
make server_run
```

Run the load test (in milmove_load_testing repo):
```
make load_test_prime
```
You can cancel this test as soon as you see the `/internal/moves/{moveId}/submit` test succeed. 

You should then log into the local office Service counselor to confirm that you can see the move that was created by the prime load test. (If you cleared your DB and only run the prime tests with one user you should only see one move on the office app.)

### DIFF
--- a/tasks/prime.py
+++ b/tasks/prime.py
@@ -231,16 +231,15 @@ class PrimeTasks(PrimeDataStorageMixin, CertTaskMixin, TaskSet):
         if resp.status_code != 200:
             self.interrupt()
 
-        logged_in_user = self.client.get(self.customer_path("/internal/users/logged_in"))
-        json_resp = logged_in_user.json()
-        service_member_id = json_resp["service_member"]["id"]
-        email = json_resp["email"]
-        user_id = json_resp["id"]
+        resp = self.client.get(self.customer_path("/internal/users/logged_in"))
+        logged_in_user = resp.json()
+        service_member_id = logged_in_user["service_member"]["id"]
+        email = logged_in_user["email"]
+        user_id = logged_in_user["id"]
 
         # Setup customer profile
-        duty_stations = self.client.get(self.customer_path("/internal/duty_stations?search=white%20sands"))
-        stations = duty_stations.json()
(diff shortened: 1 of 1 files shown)


tasks/prime.py
```diff
@@ -238,7 +238,7 @@ def on_start(self):
         user_id = json_resp["id"]
 
         # Setup customer profile
-        duty_stations = self.client.get(self.customer_path("/internal/duty_stations?search=palms"))
+        duty_stations = self.client.get(self.customer_path("/internal/duty_stations?search=white%20sands"))
```
duncan-truss, line 241: I tested this out and it works like you said.  I have a possibly hacky idea, which you should feel free to shoot down.

I see KKFA has 2 duty stations "Fort McCoy" that provides services counseling and "McConnel AFB" that would skip it and go to the TOO.  If we randomly picked one of those 2 and they are the first 2 results when searching for "McCo" then we would get seed data in the Services Counselor and TOO office queues 🤔.

Unfortunately we don't have a Services Counselor task that actually submits the move to the TOO, which is quite simple if the move is in the complete state.  We could also wait for that to be written.

duncan-truss: Nevermind I've already forgotten how the new TOO queue works, you'd have to override the postal code of the shipment as well to match.

duncan-truss (approved):

tasks/prime.py
felipe-lee: This endpoint has changed too. It's now `/duty_locations` instead of `/duty_stations`

ronaktruss: Just updated! Depending if there are additional field name changes we may have to make some additional changes.

felipe-lee (approved):

PR #111, opened 2022-01-13
MB-10534 Add docs for expected failures

felipe-lee:
## Description

This adds a section in the docs explaining how to handle expected failures.

Seconday things added:

* While adding this section, I noticed a few places in the README that I forgot to update (e.g. to use `HTTPStatus` or our new `log_response_info` and `log_response_failure` functions). 
* In writing the sample failure load test, I realized:
  * we don't quite have that test so I added it to our real load tests.
  * we had previously had a `list_moves` load test that existed to set the default mto IDs, but we got rid of it when we started creating the move in the `PrimeTasks.on_start` method. I agree we don't need it for that, but realized that we probably should still have load tests for the endpoint itself.

## Reviewer Notes

Did I miss something? Or should something be explained more?

## Setup

If you want to run the new prime tests you can run the full load tests for prime:

```sh
make load_test_prime
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10534) for this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -999,7 +999,7 @@ is the main recommendation in this repo.
 
 There may be times that you want to create a load test that you expect to fail. E.g. if you want a
 load test to cover an endpoint that will reject the request because of some error (eTag, bad data,
-etc.). If you want to do that, you can do something like this (it's a bit long because of setup):
+etc.). If you want to do that, you can do something like this:
 
 ```python
 # -*- coding: utf-8 -*-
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -993,6 +995,154 @@ thing you need to provide is the `user_type` you want to create.
...
+      log_response_info(response=resp)
+
+      if resp.status_code == HTTPStatus.OK:
+        move: JSONObject = resp.js
+      else:
+        resp.failure("Unable to get the move.")
+
+        log_response_failure(response=resp)
+
+        return
```
duncan-truss, line 1092: Because this an example, I think we should rely on just assuming there is a stored shipment we want to update to keep the focus on the failure request.  I think all of the setup detracts from the point we're trying to get across.

tasks/prime.py
```diff
@@ -378,6 +379,29 @@ def on_start(self) -> None:
         # Set local variables with needed info to create a move
         self.set_default_mto_ids(move_id)
 
+    @tag(MOVE_TASK_ORDER, "getMoves")
+    @task(2)
+    def get_moves(self) -> None:
```
duncan-truss, line 384: Thanks for adding this back, was definitely an oversight removing it when the seed setup changed.

tasks/prime.py
```diff
@@ -879,6 +903,97 @@ def update_post_counseling_information(self, overrides=None):
...
+        with self.rest(method="PATCH", url=url, data=json.dumps(payload), **request_kwargs) as resp:
+            resp: RestResponseContextManager
+
+            log_response_info(response=resp)
+
+            if resp.status_code == HTTPStatus.UNPROCESSABLE_ENTITY:
+                # Note that by default, locust would fail this request because its status code is >
+                # 400 so
+                # we need to explicitly mark it as a failure.
+                resp.success()
```
duncan-truss, line 988: I don't know if I was consistent in categorizing expected failures as successes, did you happen to make them consistent in your previous PR?

Do we already have another expected failure test for this endpoint, isn't DRAFT one of the status enum options that might be generated when making a generate_fake_request_data call already?

README.md
felipe-lee: I was torn on this because the storage as is right now causes several tests to fail so I wouldn't want someone copying this and then their test failing. I created https://dp3.atlassian.net/browse/MB-11102 to re-evaluate the storage. 

Though maybe since I ended up making this a real load test, I can shorten it in the readme and tell people to go look at the real one to see the full version.

tasks/prime.py
felipe-lee: Yeah, I fixed them all in a previous pr to count them as successes.

We have one for the support version of the endpoint, but not for the prime.

The enum excludes DRAFT: https://github.com/transcom/mymove/blob/5420f3d1a9c2b57da1c5530c3b3713fc726ad95e/swagger/prime.yaml#L2494-L2503

The support api has more statuses in the enum, but also excludes draft: https://github.com/transcom/mymove/blob/5420f3d1a9c2b57da1c5530c3b3713fc726ad95e/swagger/support.yaml#L1859-L1869

duncan-truss (approved):

README.md
ronaktruss: I'd probably lean towards a smaller scoped example so that its more focused and then linking out/referencing a specific example.

As a separate point, it might be worth adding something (we can probably add it to the story mentioned above) that in the current flow with local store we should likely be updating the store anytime a PUT, POST, or PATCH is successfully completed, so that when we try and make changes to the same object we have the updated eTags.

tasks/prime.py
felipe-lee: True 🤔 I guess in general the expected failure tests might be better to keep at a lower weight than the rest.

Though I would also say that these should be more like black-box tests in that they shouldn't know the internals of how the api validates the info, just that it gets an expected failure. Like this test should work the same whether we keep using what we are using or if we swapped out our API layer to use a different library. The swagger itself only says what the valid values should be, it's the code that the go swagger package (I think `go-openapi/swag`?) generates that does the validation for us.

README.md
felipe-lee: I agree about the setup detracting from the point of the example so I'm cutting it down now.

PR #115, opened 2022-01-21
MB-10593 Load Test Recording Tweaks

felipe-lee:
## Description

The things in this pr stemmed from the load test recording:

* Add workflow for a customer to go through on-boarding - I put this load test together for the recording, but ended up not having time to show it. I figured we should merge it in since it exists now (assuming it looks ok).
* Move locust docs to docusaurus: https://github.com/transcom/mymove-docs/pull/108

## Setup

```sh
make load_test_milmove
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10593) for this change.


tasks/milmove.py
```diff
@@ -27,11 +35,306 @@ def on_start(self):
...
+
+        if user is None:
+            return  # get_user_info will log the error
+
+        service_member_id = user["service_member"]["id"]
+
+        milmove_faker = MilMoveData()
+
+        # First let's create the profile.
+        payload = {
```
duncan-truss, line 99: I wonder if we need an includes param that lets you specify what fields to generate fake data for.  I know elsewhere we've tended to pop off fields we didn't want included after the fact but that requires you to know what other fields are in the body.

tasks/milmove.py
```diff
@@ -27,11 +35,306 @@ def on_start(self):
...
+
+                log_response_failure(response=resp, task_name=task_name)
+
+                return
+
+        # Zip code is validated as part of on-boarding.
+
+        zip_code_url, zip_code_request_kwargs = self.request_preparer.prep_internal_request(
+            endpoint=f"/rate_engine_postal_codes/{duty_location_address['postalCode']}?postal_code_type=origin",
+            endpoint_name="/rate_engine_postal_codes/{postal_code}",
```
duncan-truss, line 246: 👏 Thanks for adding this

tasks/milmove.py
```diff
@@ -27,11 +35,306 @@ def on_start(self):
...
+        # a payload.
+
+        payload = {
+            "id": service_member_id,
+            "telephone": milmove_faker.get_fake_data_for_type(data_type=DataType.PHONE),
+            "secondary_telephone": milmove_faker.get_fake_data_for_type(data_type=DataType.PHONE),
+            "personal_email": milmove_faker.get_fake_data_for_type(data_type=DataType.EMAIL),
+            "email_is_preferred": milmove_faker.get_fake_data_for_type(data_type=DataType.BOOLEAN),
+            "phone_is_preferred": milmove_faker.get_fake_data_for_type(data_type=DataType.BOOLEAN),
+        }
```
duncan-truss, line 160: I think it's fine to override and remove fields to simulate a partial patch but we're not getting the benefit of the generate_fake_request that sometimes sets optional fields and sometimes doesn't here and other places like middle_name and suffix.

tasks/milmove.py
felipe-lee: Yeah, I think it would be good to have that. I would certainly make this easier haha.

tasks/milmove.py
```diff
@@ -27,11 +35,306 @@ def on_start(self):
...
+
+        milmove_faker = MilMoveData()
+
+        # First let's create the profile.
+        payload = {
+            "id": service_member_id,
+            "affiliation": "ARMY",
+            "edipi": "1234567890",
+            "rank": "E_1",
+        }
```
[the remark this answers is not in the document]
felipe-lee: A few are based on ENUMs and I noticed the code for reading the enums and picking a random value was at the parser level so I didn't want to try to piece that together. I do think if we implemented the thing you mentioned about passing in what params you want, it would be better.

tasks/milmove.py
felipe-lee: Ah good point, I had gone back and added some of the optional fields on a second pass through, but did forget to define a value for the suffiix. And in general, yeah, this is definitely a downside to using this data generator so directly.

duncan-truss (approved):
Tested locally and customer load tests are all passing.

I think we could vary the duty station search some, I think in reality it does partial searches while the user is typing instead of a single request (it might wait for at least 2 characters though).

felipe-lee:
> Tested locally and customer load tests are all passing.
> 
> I think we could vary the duty station search some, I think in reality it does partial searches while the user is typing instead of a single request (it might wait for at least 2 characters though).

This is another good point 😅 I wasn't sure if to do that in this one or if it would be better to have a load test that focuses on testing that endpoint more thoroughly.

PR #118, opened 2022-01-27
MB-10593 Fix flaky fake middle name test

felipe-lee:
## Description

The test I wrote for the safe middle name generation is flaky because there are a lot of fake names we have without middle names. When I wrote the test and ran it, I guess I just got lucky and didn't have any of my runs select two people without middle names. This fixes it so that the check that compares them only runs if at least one of them is not empty.

## Reviewer Notes

## Setup

```sh
pytest
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10593) for this change.

### DIFF
--- a/utils/fake_data.py
+++ b/utils/fake_data.py
@@ -60,9 +60,12 @@ class MilMoveProvider(AddressProvider, DateProvider):
 
     def _set_safe_name(self):
         """
-        Randomly selects a safe full name to use.
+        Randomly selects a safe full name to use and ensure it is at least different from the
+        previous name being used.
         """
-        random_name = self.random_element(self.safe_data["names"])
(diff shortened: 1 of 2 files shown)


duncan-truss (approved):
It seems like we could still randomly get the same name chosen from the fake list right?  We don't seem to check in set_safe_name if the new random name is the same as the current one.

felipe-lee:
> It seems like we could still randomly get the same name chosen from the fake list right? We don't seem to check in set_safe_name if the new random name is the same as the current one.

That's a really good point...I'll update that. Because of how few safe names we have compared to the number of users we'll likely be making, we can't pop the names off the available list, but we can at least ensure it's not the same as the previous one. In the long term, there will be duplicates, but at least we can ensure that back-to-back ones have a different name.

PR #122, opened 2022-01-31
MB-11016 Update nix setup

felipe-lee:
## Description

Updated `NIX_PROFILE` path for multi-user installs. 

Since I don't have an M1 mac, I wasn't sure which packages needed to be updated for that so I just got newer hashes for all the packages that had one, keeping the same version as before.

Can someone with an M1 mac that uses nix check that this works?

## Setup

1. Follow [`nix` setup instructions](https://github.com/transcom/milmove_load_testing#setup-nix)

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -1,11 +1,6 @@
-# to install
-# nix-env -f nix -i
+# use ./nix/update.sh to install
 #
-# use
-#
-# https://ahobson.github.io/nix-package-search/#/search
-#
-# to find rev for specific package version
+# use https://ahobson.github.io/nix-package-search/#/search to find rev for specific package version
 
 let
   pkgs = import <nixpkgs> {};
@@ -19,7 +14,7 @@ in buildEnv {
       name = "editorconfig-core-c-0.12.1";
       url = "https://github.com/NixOS/nixpkgs/";
       ref = "refs/heads/nixpkgs-unstable";
-      rev = "c8ff5bc6f74a2960fab5ae417cd2bb055eab1002";
+      rev = "0093e4e6a9998f1ff59e79ec31f2341770b7bebb";
     }) {}).editorconfig-core-c
 
     (import (builtins.fetchGit {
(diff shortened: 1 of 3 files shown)


nix/default.nix
```diff
@@ -19,16 +14,16 @@ in buildEnv {
...
 
     (import (builtins.fetchGit {
       # Descriptive name to make the store path easier to identify
       name = "python3-3.9.6";
       url = "https://github.com/NixOS/nixpkgs/";
       ref = "refs/heads/nixpkgs-unstable";
-      rev = "c8ff5bc6f74a2960fab5ae417cd2bb055eab1002";
-    }) {}).python3
+      rev = "2f216e02f453cf64cb3b06c1e564ea53bdc01fa0";
+    }) {}).python39Full
```
ahobson, line 26: All of the git hashes in the original file work on m1. The new hashes seemingly point to an old version prior to m1 support. I think the only necessary changes are to the comment and installing the full python39

gidjin: I don't know I'm running into a different issue. I updated all the deps that had newer minor versions including python but am stuck on this one for editorconfig.

```sh
0 ❯ ./nix/update.sh
installing 'milmove-load-testing-packages'
error: cannot coerce null to a string

       at /nix/store/ri270b3pffhjbqg1x8cn1xhd5w3ij14v-editorconfig-core-c-0.12.1/pkgs/stdenv/generic/make-derivation.nix:192:19:

          191|         // (lib.optionalAttrs (!(attrs ? name) && attrs ? pname && attrs ? version)) {
          192|           name = "${attrs.pname}-${attrs.version}";
             |                   ^
          193|         } // (lib.optionalAttrs (stdenv.hostPlatform != stdenv.buildPlatform && !dontAddHostSuffix && (attrs ? name || (attrs ? pname && attrs ? version)))) {
(use '--show-trace' to show detailed location information)
```

ahobson: @gidjin That's the problem with using the hashes in this PR. The hashes that are currently on master work fine on m1 for me.

felipe-lee: @gidjin do the hashes in master work for you too? Just want to check to see if reverting them fully works for all. I've reverted all but the python one. 

As a side note, I had updated the python one to the "full" version because when I searched for hashes, 3.9.6 didn't have a regular one anymore (possibly because there are newer versions, but I'm not 100% sure).

PR #168, opened 2022-04-07
Rename duty station to location

mboren:
## Description
I renamed all instances of the term `duty station` in load testing code to `duty location`. This terminology change has already been done in the API and application code, so some load tests were failing when they tried to reference non-existent duty station keys in payloads.

Most of these places weren't generating noticeable errors, but I think that's because they were on code paths that were hitting earlier exceptions first. I haven't rigorously tested this hypothesis, but I'm confident that these keys are not used in any of our responses anymore and needed to be updated.

## Setup
Run some load tests and make sure you don't see any exceptions that mention duty stations or duty locations

```sh
make load_test_office
make load_test_prime
make load_test_milmove
```
for example, this exception that was coming from office tests should be fixed now:
image

### DIFF
--- a/tasks/milmove.py
+++ b/tasks/milmove.py
@@ -201,7 +201,7 @@ class MilMoveTasks(RestTaskSet):
         # Don't care too much about which one for this load test, so just grabbing the first one.
         origin_duty_location = duty_locations[0]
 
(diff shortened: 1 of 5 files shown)


leia-sefkin (approved):
Looks great. All tests ran without errors on the API side (except where appropriate or expected, like where we test for failures), and all the code changes look good too.

PR #178, opened 2022-04-25
[MB-12077] add multiple payment request scenario

mboren:
## Description
I added a flow that creates an HHG with a single shipment and creates two payment requests that cover all of its service items.
This is the first load testing scenario that goes through the flow up to creating a payment request, so I ran into some rough edges.
Specifically, there were a few places where return values from our API endpoints didn't match up with what the swagger spec said we'd get. I fixed some of these in [another PR](https://github.com/transcom/mymove/pull/8459), and opted out of return type checking in some other cases.

I also fixed a bug that appeared in multiple places around fetching the move from a queue.
The existing logic requested the first 100 things in the queue and checked if any of them matched the given move ID.
This doesn't really work when you get a large number of users because the move you're looking for might not appear in the first 100 queue results.
So I changed things to put the move locator into the flow context passed between steps so that each user could filter their queue for the locator.

## Reviewer Notes

## Setup
Start up the milmove server
```sh
cd ../mymove
make db_dev_reset
make db_dev_migrate
make server_run
```

Run some load tests that include this flow
```sh
locust -f locustfiles/queue.py --host local
```
Once the test is running, hop over the current ratio tab and make sure you see a non-zero percentage next to `start_hhg_multiple_payment_request_flow`.

image

Also, check the failures and exceptions tabs and make sure you're not seeing any problems.

Now let's jump into the office app and log in as a TIO in the KKFA GBLOC.
There should be exactly two payment requests for every move in the TIO queue.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-12077) for this change.
* This PR must be merged after #173

### DIFF
--- a/tasks/queue.py
+++ b/tasks/queue.py
@@ -2,7 +2,7 @@
 from locust import task, TaskSet
 
 from utils.flows import MemoryFlowQueue, QueuableFlow, WorkerQueueType
-from utils.flows.simple_hhg import SingleHHGFlow, DoubleHHGFlow, NTSFlow
+from utils.flows.simple_hhg import SingleHHGFlow, DoubleHHGFlow, NTSFlow, SingleHHGMultiplePaymentRequestFlow
 from utils.openapi_client import FlowSessionManager
 from utils.request import MilMoveRequestMixin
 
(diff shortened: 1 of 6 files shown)


ahobson (approved):
Thank you! I think this is a great step forward

PR #184, opened 2022-05-02
Use pyright to ensure typing info is correct, update deps

ahobson:
## Description

Since we are using type hints, use pyright to ensure typing info is correct as part of pre-commit

## Reviewer Notes

Minor tweak to accurately represent type info


shimonacarvalho (approved):
Having trouble with pytest in loadtesting but checked that the precommit ran smoothly.

PR #201, opened 2022-05-26
Add / update docs, allow local port to be overridden, remove reports

ahobson:
## Description

* Add more docs about the new queue approach
* Allow the local port to be overridden
* Remove reports

## Reviewer Notes

Make sure the docs make sense


shimonacarvalho (approved):
🎉

